PR #6, opened 2020-04-09
added optiongroup resource

ammokhov:
*Issue #, if available:*

*Description of changes:*

`AWS::RDS::OptionGroup`

Originally OptionGroup has [OptionConfiguration as immutable](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-optiongroup.html#cfn-rds-optiongroup-optionconfigurations), despite it allows immutable update for certain groups. 

There are two issues: cfn does not allow version downgrade on rollback.

> The aim to check the APEX version is to prevent version downgrading when doing update-rollback.
>          * 2 Downgrade will happen only when there is an APEX option to add, because only APEX has version.
>          * 3 There will be at most one APEX option in an OptionGroup
> 

Since it was implemented there appeared some other option groups that support versioning. 
Not sure what's the point of avoiding version downgrade (intentional or on rollback) it does not affect the state of target resource. 
Here is the list of all available option groups for each database engine:

- [MariaDB](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.MariaDB.Options.html)
- [Microsoft SQL Server](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.SQLServer.Options.html)
- [MySQL](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.MySQL.Options.html)
- [Oracle](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.Oracle.Options.html)

*Testing*
_Unittests_: 97% Instructions Cov., 90% Branches Cov.
_Integ Tests_: Manually tested:

- Unversioned option groups - create/update/delete
- Versioned option groups with no failures:
    - `APEX` - `4.1.1.v1` -> `4.2.6.v1` -> `4.1.1.v1`
    - `SQLT` - `12.1.160429` -> `12.2.180331` -> `12.1.160429`

Do we need to maintain the rollback limitation if so what's the rationale? 


osdrv:
Implemented in https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/35

PR #36, opened 2020-10-20
updates to make contract tests pass

carojkov:
*Description of changes:*

-- add a set of inputs for contract tests
-- add tests and implementation for ResourceModel in ReadHandler
-- remove model from result of DeleteHandler
-- add SourceDBClusterIdentifier, Engine, StorageEncrypted to createOnly list
-- give JVM more memory for SAM tests
-- update aws-cloudformation-rpdk-java-plugin to 2.0.2


aws-rds-globalcluster/src/test/java/software/amazon/rds/globalcluster/ReadHandlerTest.java
```diff
@@ -32,12 +41,6 @@
 
     private ReadHandler handler;
 
-    @AfterEach
-    public void post_execute() {
```
pabert80, line 36: Do we need to add this to all tests if we remove it from here?

carojkov: I think best to keep verify in the test, then it will be apparent which test failed.

pabert80: but you should add it to each test, with the current change we don't have the verify

pabert80 (approved):

aws-rds-globalcluster/src/test/java/software/amazon/rds/globalcluster/ReadHandlerTest.java
carojkov: that verifies the interactions with the mocks, the new tests didn't use the mocks. so actually verifying at the end of tests that don't use mocks would fails those tests.

pabert80 (approved):

aws-rds-globalcluster/src/main/java/software/amazon/rds/globalcluster/ReadHandler.java
```diff
@@ -22,9 +26,34 @@
...
+        if (cluster.hasGlobalClusterMembers()) {
+
+            List<GlobalClusterMember> globalClusterMembers = cluster.globalClusterMembers();
+
+            for (GlobalClusterMember globalClusterMember : globalClusterMembers) {
+                if (globalClusterMember.isWriter()) {
+                    builder.sourceDBClusterIdentifier(globalClusterMember.dbClusterArn());
+
+                    break;
+                }
```
ammokhov, line 53: curious if there is only one writer? if there are more than one how would we determine which one is source?

ammokhov (commented):
There are some properties that are used only on create but not included in createOnlyProperties. This is a bad customer experience as they are treated as noop and customers might get confused when engine gets updated but nothing happens

aws-rds-globalcluster/src/main/java/software/amazon/rds/globalcluster/ReadHandler.java
carojkov: @pabert80 @Dmytrok-amz could you guys confirm that?

pabert80: There is only one writer in Global Databases at the time

ammokhov (approved):

Issue #43, opened 2020-11-11
Include DbClusterResourceId as accessible attribute of DbCluster


ghost:
(This issue duplicates https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/192, as I'm interpreting the contribution guidelines in this way; if that is incorrect, please let me know.)

This attribute is needed for some specific circumstances, and cannot currently be retrieved through GetAttr.

osdrv:
This attribute was added to the resource and rolled out in all AWS regions. Closing this issue for now.

PR #44, opened 2020-11-11
Add DBClusterResourceId as read-only attribute (#43)

tjp502:
Issue #43 

Changes as title; this is a very simple change. Currently has no test associated with it, as I couldn't see any place to add one.

### DIFF
--- a/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/ReadHandler.java
+++ b/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/ReadHandler.java
@@ -39,6 +39,7 @@ public class ReadHandler extends BaseHandlerStd {
                                     .port(targetDBCluster.port().toString()).build())
                             .readEndpoint(ReadEndpoint.builder()
                                     .address(targetDBCluster.readerEndpoint()).build())
+                            .dBClusterResourceId(targetDBCluster.dbClusterResourceId())
 
                             .associatedRoles(targetDBCluster.associatedRoles().stream().map(roleTransform).collect(Collectors.toList()))
                             .availabilityZones(targetDBCluster.availabilityZones())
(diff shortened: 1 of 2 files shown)


aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/ReadHandler.java
```diff
@@ -39,6 +39,7 @@
                                     .port(targetDBCluster.port().toString()).build())
                             .readEndpoint(ReadEndpoint.builder()
                                     .address(targetDBCluster.readerEndpoint()).build())
+                            .dBClusterResourceId(targetDBCluster.dbClusterResourceId())
```
ammokhov, line 42: thanks for your contribution. This is a good start but Create/Update/Read handlers should output same set of properties

tjp502: This is a read-only property, so I think that means it only needs to be in the Read handler; as far as I know there's no way to set this on Create or Update.

ammokhov: > The model MUST contain all properties that have values, including any properties that have default values, and any readOnlyProperties as defined in the resource schema

[contract agreement](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html#resource-type-test-contract-create)

read only properties are mainly used for GetAtt or something that could not be provided via template but we require consistent output between handlers

tjp502: I think I understand, but I don't see where the existing read-only properties are being set in the CreateHandler or UpdateHandler code paths. :-( Can you point me to an example of this in practice, so I can use that as a model?

hesiod-au:
It looks to me like the contributor has answered the question;  Are there any further issues, what can be done to get this improvement into the code? Please see https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/192 there are certain configurations that NEED this property available in CloudFormation and it has been requested for years

iainb123:
Hi @ammokhov what are the next steps here to get this released?

Issue #40, opened 2020-11-05
[DBInstance] Implement AWS::RDS::DBInstance


osdrv:
RDS WS API team owns the resource

PR #46, opened 2020-12-01
Implement AWS::RDS::DBInstance resource

osdrv:
This PR provides a complete AWS::RDS::DBInstance resource handler implementation.

Resource specification:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html

Original issue: #40 


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -0,0 +1,220 @@
...
+                    return isDbInstanceStabilized(proxyInvocation, resourceModel, DBInstanceStatus.Available);
+                })
+                .progress();
+    }
+
+    protected ListTagsForResourceRequest listTagsForResourceRequest(final String dbInstanceArn) {
+        return ListTagsForResourceRequest.builder()
+                .resourceName(dbInstanceArn)
+                .build();
+    }
```
osdrv, line 80: TODO: move this to Translator

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/ListHandler.java
```diff
@@ -0,0 +1,42 @@
...
+                Translator.describeDbInstancesRequest(request.getNextToken()),
+                proxyClient.client()::describeDBInstances
+        );
+
+        return ProgressEvent.<ResourceModel, CallbackContext>builder()
+                .resourceModels(describeDBInstancesResponse.dbInstances()
+                        .stream()
+                        .map(dbInstance -> ResourceModel
+                                    .builder()
+                                    .dBInstanceIdentifier(dbInstance.dbInstanceIdentifier())
```
osdrv, line 34: Is it really sufficient?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -0,0 +1,98 @@
...
+    protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
+        final AmazonWebServicesClientProxy proxy,
+        final ResourceHandlerRequest<ResourceModel> request,
+        final CallbackContext callbackContext,
+        final ProxyClient<RdsClient> proxyClient,
+        final Logger logger) {
+
+        final ResourceModel model = request.getDesiredResourceState();
+
+        if (StringUtils.isEmpty(model.getDBInstanceIdentifier())) {
```
osdrv, line 23: TODO: move this to ProgressEvent.progress().then(), otherwise a new name would be generated every time create request fails, which is fairly dangerous.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/UpdateHandler.java
```diff
@@ -0,0 +1,183 @@
...
+                .then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
+    }
+
+    private ProgressEvent<ResourceModel, CallbackContext> updateTags(
+            final AmazonWebServicesClientProxy proxy,
+            final ProxyClient<RdsClient> proxyClient,
+            ProgressEvent<ResourceModel, CallbackContext> progress,
+            final Collection<Tag> desiredTags
+    ) {
+        return proxy.initiate("rds-tag-db-instance", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```
osdrv, line 71: unconventional callGraph name: should be `"rds::tag-db-instance"`

aws-rds-dbinstance/src/test/java/software/amazon/rds/dbinstance/ListHandlerTest.java
```diff
@@ -0,0 +1,88 @@
...
+        when(proxyClient.client().describeDBInstances(any(DescribeDbInstancesRequest.class))).thenReturn(describeDbInstanceResponse);
+
+        final ResourceModel model = ResourceModel.builder().build();
+
+        final ResourceModel expectedModel = ResourceModel.builder()
+                .dBInstanceIdentifier(DB_INSTANCE_IDENTIFIER)
+                .build();
+
+        final ResourceHandlerRequest<ResourceModel> request = ResourceHandlerRequest.<ResourceModel>builder()
+                .desiredResourceState(model)
```
osdrv, line 70: is providing a desired resource state for ListHandler really needed?

aws-rds-dbinstance/src/test/java/software/amazon/rds/dbinstance/ListHandlerTest.java
```diff
@@ -0,0 +1,88 @@
...
+    final String DB_INSTANCE_IDENTIFIER = "test-db-instance-identifier";
+    final String DESCRIBE_DB_INSTANCES_MARKER = "test-describe-db-instances-marker";
+
+    @BeforeEach
+    public void setup() {
+        handler = new ListHandler();
+        rdsClient = mock(RdsClient.class);
+        proxy = new AmazonWebServicesClientProxy(logger, MOCK_CREDENTIALS, () -> Duration.ofSeconds(600).toMillis());
+        proxyClient = MOCK_PROXY(proxy, rdsClient);
+    }
```
osdrv, line 51: missing AfterEach

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/UpdateHandler.java
```diff
@@ -0,0 +1,183 @@
...
+                            .translateToServiceRequest(resourceModel -> Translator.modifyDbInstanceRequest(previousModel, desiredModel))
+                            .backoffDelay(BACKOFF_STRATEGY)
+                            .makeServiceCall((modifyRequest, proxyInvocation) -> {
+                                return proxyInvocation.injectCredentialsAndInvokeV2(
+                                        modifyRequest,
+                                        proxyInvocation.client()::modifyDBInstance
+                                );
+                            })
+                            .done((modifyRequest, modifyResponse, proxyInvocation, resourceModel, context) -> {
+                                context.setModified(true);
```
osdrv, line 55: No need to update the context if the test suite does not use this transition assessment

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
+{
```
moataz-mhmd, line 1: Could you elaborate why `EnableCustomerOwnedIp , NcharCharacterSetName , TdeCredentialArn` parameters are not included in the schema? 
reference: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+      "description": "For supported engines, indicates that the DB instance should be associated with the specified character set."
+    },
+    "CopyTagsToSnapshot": {
+      "type": "boolean",
+      "description": "A value that indicates whether to copy tags from the DB instance to snapshots of the DB instance. By default, tags are not copied."
+    },
+    "DBClusterIdentifier": {
+      "type": "string",
+      "description": "The identifier of the DB cluster that the instance will belong to.",
+      "pattern": "^[a-zA-Z]{1}(?:-?[a-zA-Z0-9]){0,62}$"
```
moataz-mhmd, line 113: As it mentioned here:https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html under *DBInstanceIdentifier* parameter. The maximum length is 63 letters
This regex will accept string with length up to 126. For Example: this is a test case will pass this pattern

a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a-a

I would suggest you use this to avoid complex patterns and simplicity 
```
"pattern": "^[a-zA-Z](?:-?[a-zA-Z0-9])$",
"minLength": 1,
"maxLength": 63
```

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+    },
+    "DBInstanceClass": {
+      "type": "string",
+      "description": "The compute and memory capacity of the DB instance, for example, db.m4.large. Not all DB instance classes are available in all AWS Regions, or for all database engines."
+    },
+    "DBInstanceIdentifier": {
+      "type": "string",
+      "pattern": "^[a-zA-Z]{1}(?:-?[a-zA-Z0-9]){0,62}$",
```
moataz-mhmd, line 121: Same as DBClusterIdentifier

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+      "description": "A name for the DB instance. If you specify a name, AWS CloudFormation converts it to lowercase. If you don't specify a name, AWS CloudFormation generates a unique physical ID and uses that ID for the DB instance."
+    },
+    "DBName": {
+      "type": "string",
+      "pattern": "^[a-zA-Z][a-zA-Z0-9]{0,63}$",
```
moataz-mhmd, line 126: As it mentioned here: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html under *DBName* section. Underscores  are valid input in *[MySQL, MariaDB, PostgreSQL]* engine cases. In case of using *Amazon Aurora PostgreSQL* customer can start *DBName* with underscore. This pattern will reject these scenario.

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+      "description": "The meaning of this parameter differs according to the database engine you use."
+    },
+    "DBParameterGroupName": {
```
moataz-mhmd, line 129: Should we add a pattern and minLength and maxLength? as It mentioned here in https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html as constrains under *DBParameterGroupName* like:
```
"pattern": "^[a-zA-Z]{1}(?:-?[a-zA-Z0-9])*$",
minLength: 1
maxLength: 255
```

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+        "aurora-mysql",
+        "aurora-postgresql",
+        "mariadb",
+        "mysql",
+        "oracle-ee",
+        "oracle-se2",
+        "oracle-se1",
+        "oracle-se",
+        "postgres",
+        "postgresql-license",
```
moataz-mhmd, line 198: "postgresql-license",

*aws-rds-dbinstance:198*

Could you elaborate this version name or provide a doc? 
I didn’t find it under engine valid values here : https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+      ],
+      "description": "The name of the database engine that you want to use for this DB instance."
+    },
+    "EngineVersion": {
+      "type": "string",
+      "description": "The version number of the database engine to use."
+    },
+    "Iops": {
+      "type": "integer",
+      "minimum": 1000,
```
moataz-mhmd, line 212: Should we specify a maximum 80000 as it mentioned in doc?
Reference: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html#USER_PIOPS

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+        "bring-your-own-license",
+        "general-public-license",
+        "license-included",
+        "postgresql-license"
+      ],
+      "description": "License model information for this DB instance."
+    },
+    "MasterUsername": {
+      "type": "string",
+      "pattern": "^[a-zA-Z][a-zA-Z0-9]{0,15}$",
```
moataz-mhmd, line 231: This pattern will refuse a valid input in case of these engines

* *Microsoft SQL Server* which length can go up to 128
* *Oracle* length up to 30
* *PostgreSQL* length up to 63

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+      "description": "The master user name for the DB instance."
+    },
+    "MasterUserPassword": {
```
moataz-mhmd, line 234: Should we add some specification as it mentioned in docs? like
```
"pattern": "^[^\"/@]+$",
"minLength": 8,
"maxLength": 128
```
reference: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,431 @@
...
+      "type": "string",
+      "description": "The ARN of the AWS Key Management Service (AWS KMS) master key that's used to encrypt the DB instance."
+    },
+    "LicenseModel": {
+      "type": "string",
+      "enum": [
+        "bring-your-own-license",
+        "general-public-license",
+        "license-included",
+        "postgresql-license"
```
moataz-mhmd, line 225: Could you elaborate this LicenseModel or provide a doc? 
I didn’t find it under LicenseModel valid values here : https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBInstance.html

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -0,0 +1,439 @@
...
+import software.amazon.awssdk.services.rds.model.DeleteDbInstanceRequest;
+import software.amazon.awssdk.services.rds.model.DescribeDbInstancesRequest;
+import software.amazon.awssdk.services.rds.model.ListTagsForResourceRequest;
+import software.amazon.awssdk.services.rds.model.ModifyDbInstanceRequest;
+import software.amazon.awssdk.services.rds.model.RemoveRoleFromDbInstanceRequest;
+import software.amazon.awssdk.services.rds.model.RemoveTagsFromResourceRequest;
+import software.amazon.awssdk.services.rds.model.RestoreDbInstanceFromDbSnapshotRequest;
+import software.amazon.awssdk.utils.StringUtils;
+
+public class Translator {
```
moataz-mhmd, line 29: Just a reminder to update this class if you decide to add any parameter to schema.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
```diff
@@ -0,0 +1,73 @@
...
+                            deleteRequest,
+                            proxyInvocation.client()::deleteDBInstance
+                    );
+                    callbackContext.setDeleting(true);
+                    return response;
+                })
+                .stabilize((deleteRequest, deleteResponse, proxyInvocation, model, context) -> {
+                    return isDbInstanceDeleted(proxyInvocation, model);
+                })
+                .handleError((deleteRequest, exception, client, model, context) -> {
```
moataz-mhmd, line 63: One of expected exceptions that deleteDBInstance can throw `DbSnapshotAlreadyExistsException` 
Should we handle it? I am not sure. Like
```
if (exception instanceof DbSnapshotAlreadyExistsException)
    return ProgressEvent.defaultFailureHandler(exception, HandlerErrorCode.AlreadyExists);
}
```

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -0,0 +1,439 @@
...
+
+    public static ResourceModel translateDbInstanceFromSdk(final software.amazon.awssdk.services.rds.model.DBInstance dbInstance, final List<software.amazon.awssdk.services.rds.model.Tag> tags) {
+
+        final String dbParameterGroupName = Optional.ofNullable(dbInstance.dbParameterGroups())
+                .orElse(Collections.emptyList())
+                .stream()
+                .map(Translator::translateDBParameterGroupFromSdk)
+                .findFirst().orElse(null);
+
+        // https://rds-jira.amazon.com/browse/WS-293
```
moataz-mhmd, line 280: Should we remove this comment as this tool is for internal use only? And add explanation why we are doing this?

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: None of these are present in the current spec: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html
We can take a note to add support of these in the future as a separate PR.

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: Amazing catch! Regarding using a pattern together with length limiters: the cfn tool warns about it. Idiomatically there is a chance of a conflict of these 3 constraints, so the best practice is to either specify min/max or a pattern.

osdrv: About the pattern itself: we use it in multiple places, e.g.:
* https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbcluster/aws-rds-dbcluster.json#L49
* https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-globalcluster/aws-rds-globalcluster.json#L26
I believe it is kept this way for the sake of simplicity (the no-dual hyphen means a look-ahead regex move with back-referencing).
I believe we should not treat this validation as the last border control to the resource definition. WS will do per-instance validation, in this case a CFN user will see an error once the request is sent to the backend. In other words, the CFN validation could be much more relaxed and not cover all possible cases (which might be pretty flexible and dynamic on the server side).

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: This came as a surprise to me when I was running contract tests.

If you run a basic create-db-instance for a postgres db:
```
aws rds --region us-east-1 create-db-instance --db-name mydb --db-instance-identifier mydbinstance --allocated-storage 20 --engine postgres --db-instance-class db.t3.micro --master-username=<username> --master-user-password=<password>
```
The output would contain:
```
"LicenseModel": "postgresql-license",
```

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: It is 256k for Oracle. I'd say let's leave this out as it is engine-specific and would be rejected by WS if exceeds the limit anyways.

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: Good catch, I must keep this filter broader than WS. To be fixed.

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: This came out of the blue while running contract tests:

`aws rds create-db-instance --db-name <db-name> --db-instance-identifier <db-instance-identifier> --allocated-storage 20 --engine postgres --db-instance-class db.t3.micro --master-username=<master-username> --master-user-password=<master-user-password>`

The output would contain:
```
"LicenseModel": "postgresql-license",
```

Strictly speaking, I have some doubts about this attribute: it is extremely fragile to adding new licenses and customers barely use it. Keeping in for the backwards compatibility.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
osdrv: This is a great question and it puzzled me for a while. I believe the intent here is to wrap RDS-specific 404 to CFN 404, which is expected to be handled by Uluru itself. The details of the failure would be available to the customer anyways. If I understand this part correctly, the intent is to distinguish retryable errors and dead-end ones. This must be taken through the prism of CFN peculiarity: a retryable from CLI/SDK perspective is unlike CFN retryable in a way: 4xx is not supposed to be retryable by the SDK because there is a conflicting state which the customer is supposed to resolve and then retry explicitly. 5xx is a retryable thing. In the CFN case the conflict is expecting a side effect from the customer, but the retry process is implicit unlike the previous one. This means that a 4xx error could expect the customer to eventually resolve the conflict and the automation will kick in. Except this 404: there is nothing to expect.
Having that said, I'm leaning towards handling 404 only in this context, hope this makes sense.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -0,0 +1,213 @@
...
+import software.amazon.cloudformation.exceptions.CfnNotFoundException;
+import software.amazon.cloudformation.exceptions.CfnNotStabilizedException;
+import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
+import software.amazon.cloudformation.proxy.Logger;
+import software.amazon.cloudformation.proxy.ProgressEvent;
+import software.amazon.cloudformation.proxy.ProxyClient;
+import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
+import software.amazon.cloudformation.proxy.delay.Constant;
+
+public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
```
moataz-mhmd, line 28: Should we gather all error handling here and process expected exception?
I think that will be easy to maintain and make cloud formation more aware of exceptions for any future feature [Like giving insights to customer(especially with automation scripts) for throttled requests or invalid parameter exceptions] and customer still  RDS exception anyway.

For sure It is not mandatory and code will work without it. But is it nice to have?

To relate, please take a look how `aws ApiGateway resource`  do it: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-apigateway/blob/5751e9513c14c1248f1b3534efb5d7358887c837/resource/src/main/java/software/amazon/apigateway/resource/BaseHandlerStd.java#L58

osdrv: Good point, got it implemented here: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/46/files#diff-3429914b497911f240a27f529350ff58bb2e18db96883b9c13f609194ae24c57R217-R241

moataz-mhmd (dismissed):

osdrv:
The failing pipeline build is blocked by https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/51/

sungkkim (commented):
I don't see Endpoint.Address and Endpoint.Port in the resource schema. These need to be marked as read-only properties to continue supporting GetAtt

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -0,0 +1,125 @@
...
+                .then(progress -> {
+                    if (isReadReplica(progress.getResourceModel())) {
+                        return initiateCreateDbInstanceReadReplicaRequest(proxy, proxyClient, progress);
+                    } else if (isRestoreFromSnapshot(progress.getResourceModel())) {
+                        return initiateRestoreDbInstanceFromSnapshotRequest(proxy, proxyClient, progress);
+                    }
+                    return initiateCreateDbInstanceRequest(proxy, proxyClient, progress);
+                })
+                .then(progress -> waitForDbInstanceAvailableStatus(proxy, proxyClient, progress))
+                .then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
```
sungkkim, line 52: Do we need to reboot the instance if it has a parameter group name? Also, are we calling `addRoleToDBInstance` from the create handler somewhere?

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+    },
+    "DBSecurityGroups": {
+      "type": "array",
+      "uniqueItems": true,
+      "items": {
+        "type": "string"
+      },
+      "description": "A list of the DB security groups to assign to the DB instance. The list can include both the name of existing DB security groups or references to AWS::RDS::DBSecurityGroup resources created in the template."
+    },
+    "DBSnapshotIdentifier": {
```
ammokhov, line 142: this should be modeled via conditional operators - `oneOf`, `allOf`, `anyOf`
```
If you specify the DBSnapshotIdentifier property to restore a DB instance (as opposed to specifying it for DB instance updates), then don't specify the following properties:
    AllocatedStorage
    BackupRetentionPeriod
    CharacterSetName
    DBClusterIdentifier
    DBName
    DBSecurityGroups
    DeleteAutomatedBackups
    EnablePerformanceInsights
    EngineVersion
    KmsKeyId
    MasterUsername
    MasterUserPassword
    MaxAllocatedStorage
    MonitoringInterval
    MonitoringRoleArn
    PerformanceInsightsKMSKeyId
    PerformanceInsightsRetentionPeriod
    PreferredBackupWindow
    PreferredMaintenanceWindow
    PromotionTier
    SourceDBInstanceIdentifier
    SourceRegion
    StorageEncrypted
    Timezone
```

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+      "type": "string",
+      "description": "The Active Directory directory ID to create the DB instance in. Currently, only MySQL, Microsoft SQL Server, Oracle, and PostgreSQL DB instances can be created in an Active Directory Domain."
+    },
+    "DomainIAMRoleName": {
+      "type": "string",
+      "description": "Specify the name of the IAM role to be used when making API calls to the Directory Service."
+    },
+    "EnableCloudwatchLogsExports": {
+      "type": "array",
+      "uniqueItems": true,
```
ammokhov, line 170: list instead of set

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+      "type": "string",
+      "description": "The daily time range during which automated backups are created if automated backups are enabled, using the BackupRetentionPeriod parameter."
+    },
+    "PreferredMaintenanceWindow": {
+      "type": "string",
+      "description": "he weekly time range during which system maintenance can occur, in Universal Coordinated Time (UTC)."
+    },
+    "ProcessorFeatures": {
+      "type": "array",
+      "uniqueItems": true,
```
ammokhov, line 293: list instead of set

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+        "type": "string"
+      },
+      "description": "A list of the VPC security group IDs to assign to the DB instance. The list can include both the physical IDs of existing VPC security groups and references to AWS::EC2::SecurityGroup resources created in the template."
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "DBInstanceClass"
+  ],
+  "createOnlyProperties": [
```
ammokhov, line 360: you should specify `conditionalCreateOnlyProperties` - [Adding conditionallyCreateOnlyProperties to metaschema (#121)](https://github.com/aws-cloudformation/cloudformation-resource-schema/commit/7d82244d54d4178b9a0f5688c39d1050b0ff3cb4)

```
"conditionalCreateOnlyProperties": [
  "/properties/AutoMinorVersionUpgrade",
  "/properties/BackupRetentionPeriod",
  "/properties/DBParameterGroupName",
  "/properties/Engine",
  "/properties/EngineVersion",
  "/properties/MonitoringInterval",
  "/properties/PerformanceInsightsKMSKeyId",
  "/properties/PreferredMaintenanceWindow",
  "/properties/StorageType"
]

```

ammokhov (changes requested):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/ReadHandler.java
```diff
@@ -0,0 +1,36 @@
...
+
+        return proxy.initiate("rds::describe-db-instance", proxyClient, request.getDesiredResourceState(), callbackContext)
+                .translateToServiceRequest(Translator::describeDbInstancesRequest)
+                .makeServiceCall((describeRequest, proxyInvocation) -> proxyInvocation.injectCredentialsAndInvokeV2(
+                        describeRequest,
+                        proxyInvocation.client()::describeDBInstances
+                ))
+                .done((describeRequest, describeResponse, proxyInvocation, model, context) -> {
+                    final DBInstance dbInstance = describeResponse.dbInstances().stream().findFirst().get();
+                    final List<Tag> tags = listTagsSoftFailOnAccessDenied(proxyInvocation, dbInstance.dbInstanceArn());
```
moataz-mhmd, line 29: Sorry for late comment. describe response already have tag list in it.
Could you elaborate why we made anther `ListTagForResource` request instead of fetching them from response?
i.e: `dbInstance.tagList()`

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+        "rds:RemoveTagsFromResource",
+        "rds:AddTagsToResource"
+      ]
+    },
+    "delete": {
+      "permissions": [
+        "rds:DescribeDBInstances",
+        "rds:DeleteDBInstance",
+        "rds:ListTagsForResource",
+        "rds:RemoveTagsFromResource"
```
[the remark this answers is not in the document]
osdrv: Removed everywhere except `ReadHandler`

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+    "/properties/DBInstanceIdentifier"
+  ],
+  "handlers": {
+    "create": {
+      "permissions": [
+        "rds:CreateDBInstance",
+        "rds:ModifyDBInstance",
+        "rds:DescribeDBInstances",
+        "rds:AddRoleToDBInstance",
+        "rds:ListTagsForResource"
```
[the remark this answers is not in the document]
osdrv: Good point. Fixed.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
osdrv: This part was completely missing in the initial impl, thanks very much for finding it!

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+      },
+      "required": [
+        "Key",
+        "Value"
+      ]
+    }
+  },
+  "properties": {
+    "AllocatedStorage": {
+      "type": "integer",
```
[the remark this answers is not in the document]
osdrv: Fixed. Thank you!

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+  "definitions": {
+    "DBInstanceRole": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "FeatureName": {
+          "type": "string",
+          "description": "The name of the feature associated with the AWS Identity and Access Management (IAM) role. IAM roles that are associated with a DB instance grant permission for the DB instance to access other AWS services on your behalf."
+        },
+        "RoleArn": {
```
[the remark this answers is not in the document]
osdrv: As far as I can see, we expose no specific ARN for this attribute. Neither public nor internal contract declares it. Proposing to keep it so.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/ReadHandler.java
osdrv: Good catch, thanks very much!

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: Fixed

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: Added. Thank you.

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+      "type": "integer",
+      "description": "The amount of storage (in gigabytes) to be initially allocated for the database instance."
+    },
+    "AllowMajorVersionUpgrade": {
+      "type": "boolean",
+      "description": "A value that indicates whether major version upgrades are allowed. Changing this parameter doesn't result in an outage and the change is asynchronously applied as soon as possible."
+    },
+    "AssociatedRoles": {
+      "type": "array",
+      "uniqueItems": true,
```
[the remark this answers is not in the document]
osdrv: Fixed

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+      "description": "The compute and memory capacity of the DB instance, for example, db.m4.large. Not all DB instance classes are available in all AWS Regions, or for all database engines."
+    },
+    "DBInstanceIdentifier": {
+      "type": "string",
+      "pattern": "^[a-zA-Z]{1}(?:-?[a-zA-Z0-9]){0,62}$",
+      "description": "A name for the DB instance. If you specify a name, AWS CloudFormation converts it to lowercase. If you don't specify a name, AWS CloudFormation generates a unique physical ID and uses that ID for the DB instance."
+    },
+    "DBName": {
+      "type": "string",
+      "pattern": "^[_a-zA-Z][a-zA-Z0-9]{0,63}$",
```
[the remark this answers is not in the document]
osdrv: It looks like the most deliberate unification of https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-dbname

aws-rds-dbinstance/aws-rds-dbinstance.json
osdrv: @ammokhov could you please elaborate how a negation could be implemented with the current schema?

I tried something like:
```json
{
  "not": {
    "required": [
      "AllocatedStorage",
      "BackupRetentionPeriod",
      ...
      "SourceRegion",
      "StorageEncrypted",
      "Timezone"
    ]
  }
}
```
But this conflicts with the [schema definition](https://schema.cloudformation.us-east-1.amazonaws.com/provider.definition.schema.v1.json).

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,427 @@
...
+    "DBInstanceClass": {
+      "type": "string",
+      "description": "The compute and memory capacity of the DB instance, for example, db.m4.large. Not all DB instance classes are available in all AWS Regions, or for all database engines."
+    },
+    "DBInstanceIdentifier": {
+      "type": "string",
+      "pattern": "^[a-zA-Z]{1}(?:-?[a-zA-Z0-9]){0,62}$",
+      "description": "A name for the DB instance. If you specify a name, AWS CloudFormation converts it to lowercase. If you don't specify a name, AWS CloudFormation generates a unique physical ID and uses that ID for the DB instance."
+    },
+    "DBName": {
```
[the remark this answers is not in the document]
osdrv: Let's keep this thread below as it is concerning the same subject.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
```diff
@@ -0,0 +1,69 @@
...
+            final CallbackContext callbackContext,
+            final ProxyClient<RdsClient> rdsProxyClient,
+            final ProxyClient<Ec2Client> ec2ProxyClient,
+            final Logger logger
+    ) {
+        final ResourceModel resourceModel = request.getDesiredResourceState();
+        String snapshotIdentifier = null;
+        // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html
+        // For AWS::RDS::DBInstance resources that don't specify the DBClusterIdentifier property, the default policy is Snapshot.
+        if (BooleanUtils.isNotFalse(request.getSnapshotRequested())) {
```
sungkkim, line 35: Nice

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,476 @@
...
+        "type": "string"
+      },
+      "description": "A list of the VPC security group IDs to assign to the DB instance. The list can include both the physical IDs of existing VPC security groups and references to AWS::EC2::SecurityGroup resources created in the template."
+    }
+  },
+  "additionalProperties": false,
+  "oneOf": [
+    {
+      "required": [
+        "DBInstanceIdentifier",
```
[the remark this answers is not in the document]
osdrv: Fixed. Thank you.

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,476 @@
...
+    "/properties/KmsKeyId",
+    "/properties/MasterUsername",
+    "/properties/Port",
+    "/properties/PubliclyAccessible",
+    "/properties/SourceDBInstanceIdentifier",
+    "/properties/SourceRegion",
+    "/properties/StorageEncrypted",
+    "/properties/Timezone"
+  ],
+  "conditionalCreateOnlyProperties": [
```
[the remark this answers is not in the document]
osdrv: Added. Thank you.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
osdrv: <3

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,476 @@
...
+    "/properties/StorageType"
+  ],
+  "writeOnlyProperties": [
+    "/properties/MasterUserPassword",
+    "/properties/DBSnapshotIdentifier",
+    "/properties/SourceDBInstanceIdentifier",
+    "/properties/SourceRegion"
+  ],
+  "readOnlyProperties": [
+    "/properties/Endpoint",
```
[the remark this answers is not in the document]
osdrv: Removed `Endpoint` from this list. The attributes remain in the list.

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,476 @@
...
+      "type": "string",
+      "description": "The time zone of the DB instance. The time zone parameter is currently supported only by Microsoft SQL Server."
+    },
+    "UseDefaultProcessorFeatures": {
+      "type": "boolean",
+      "description": "A value that indicates whether the DB instance class of the DB instance uses its default processor features."
+    },
+    "VPCSecurityGroups": {
+      "type": "array",
+      "items": {
```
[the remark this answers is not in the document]
osdrv: Fixed both. Thank you.

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -0,0 +1,476 @@
+{
+  "typeName": "AWS::RDS::DbInstance",
```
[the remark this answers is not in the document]
osdrv: Done.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -0,0 +1,299 @@
...
+                        exception instanceof StorageQuotaExceededException
+        ) {
+            return ProgressEvent.defaultFailureHandler(exception, HandlerErrorCode.ServiceLimitExceeded);
+        } else if (
+                exception instanceof InvalidDbInstanceStateException ||
+                        exception instanceof InvalidDbClusterStateException ||
+                        exception instanceof DbSnapshotAlreadyExistsException ||
+                        exception instanceof DbUpgradeDependencyFailureException
+        ) {
+            return ProgressEvent.defaultFailureHandler(exception, HandlerErrorCode.ResourceConflict);
```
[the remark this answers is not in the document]
osdrv: It seems we define no dedicated error type for this case in RDS schema :-(

ammokhov (approved):

ammokhov:
enormous PR! overall implementation looks good. I might have missed something but at this point we should merge and do iterative fixes if necessary

moataz-mhmd (approved):
Looks good to me. I am also agree with Anton. We can start with resource and do iterative fixes in the future if any issue appeared

PR #54, opened 2021-05-04
AWS::RDS::OptionGroup supports system tags

osdrv:
This commit adds a support for system tags. Prior to that, OptionGroup
was supporting resource-level tags only. RDS API makes no distinction
between system- and resource-level tags, so the tags are merged into a
single collection before invoking the corresponding RDS API call.

Apart from this change, OptionGroup drops a support for tagging
soft-failing. As it was concluded internally, soft failing on tagging is
considered to be a deprecated mode, therefore it is not making its way
to the new implementation.

AWS SDK version was bumped to 2.16.54.

### DIFF
--- a/aws-rds-optiongroup/.rpdk-config
+++ b/aws-rds-optiongroup/.rpdk-config
@@ -1,5 +1,4 @@
 {
-    "artifact_type": "RESOURCE",
     "typeName": "AWS::RDS::OptionGroup",
     "language": "java",
     "runtime": "java8",
@@ -14,6 +13,5 @@
         ],
         "codegen_template_path": "guided_aws",
         "protocolVersion": "2.0.0"
-    },
-    "executableEntrypoint": "software.amazon.rds.optiongroup.HandlerWrapperExecutable"
+    }
 }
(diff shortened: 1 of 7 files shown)


carojkov (approved):

moataz-mhmd:
I have a general question, In case that customer stack doesn't contain tags and customer doesn't have the permission to attach tags but CFN will attach system tags anyway to resources in this stack as mentioned here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html 
> In addition to any tags you define, CloudFormation automatically creates the following stack-level tags with the prefix aws::
    aws:cloudformation:logical-id
    aws:cloudformation:stack-id
    aws:cloudformation:stack-name
    All stack-level tags, including automatically created tags, are propagated to resources that CloudFormation supports. Currently, tags aren't propagated to Amazon EBS volumes that are created from block device mappings. 
questions are:
1. Is stack going to succeed even customer doesn't have the permission to attach these system tags? If yes, is it acceptable result to have the resource without these default system tags?

moataz-mhmd (approved):

PR #60, opened 2021-07-28
Changes to meet contract tests

moataz-mhmd:
All contract test are passing

### DIFF
--- a/aws-rds-dbsubnetgroup/src/main/java/software/amazon/rds/dbsubnetgroup/BaseHandlerStd.java
+++ b/aws-rds-dbsubnetgroup/src/main/java/software/amazon/rds/dbsubnetgroup/BaseHandlerStd.java
@@ -1,94 +1,112 @@
 package software.amazon.rds.dbsubnetgroup;
 
+import static software.amazon.rds.dbsubnetgroup.Translator.addTagsToResourceRequest;
+import static software.amazon.rds.dbsubnetgroup.Translator.listTagsForResourceRequest;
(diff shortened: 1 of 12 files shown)


osdrv (dismissed):

aws-rds-dbsubnetgroup/src/main/java/software/amazon/rds/dbsubnetgroup/BaseHandlerStd.java
```diff
@@ -56,6 +59,31 @@ protected boolean isStabilized(final ResourceModel model, final ProxyClient<RdsC
       return status.equals(DB_SUBNET_GROUP_STATUS_COMPLETE);
     }
 
+    public ProgressEvent<ResourceModel, CallbackContext> handleException(final Exception e) {
+        if (
```
carojkov, line 63: I am not sure if this is the formatting we use. I thought we had something like this

```
if (e instanceof Blah) {
    return blah;
}

```
expression in if or else if - in one line
4 spaces indentation.

carojkov (changes requested):

aws-rds-dbsubnetgroup/src/main/java/software/amazon/rds/dbsubnetgroup/BaseHandlerStd.java
moataz-mhmd: Done

osdrv (approved):

carojkov (approved):

PR #92, opened 2021-12-07
Unset DBInstance MaxAllocatedStorage

osdrv:
This commit addresses unsetting `MaxAllocatedStorage` property upon a DBInstance update.

Given a CFN template (state A):
```yaml
Resources:
  DBInstance:
    Type: AWS::RDS::DBInstance
    Properties:
      AllocatedStorage: 100
      ...
```
Provided an update (state B):
```yaml
Resources:
  DBInstance:
    Type: AWS::RDS::DBInstance
    Properties:
      AllocatedStorage: 100
      MaxAllocatedStorage: 1000
      ...
```
Transition to state B will set `MaxAllocatedStorage` to 1000 and enable the autoscaling.

If `MaxAllowedStorage` is unset later in the template (transition back to state A or its equivalent), the current implementation will keep the property in place rather than unsetting it.

RDS API does not have an explicit request parameter to unset this property. The way to get around it is to set `MaxAllocatedStorage` to the current `AllocatedStorage`.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/UpdateHandler.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/UpdateHandler.java
@@ -89,6 +89,12 @@ public class UpdateHandler extends BaseHandlerStd {
                     }
                     return progress;
                 })
+                .then(progress -> {
+                    if (shouldUnsetMaxAllocatedStorage(request)) {
+                        return unsetMaxAllocatedStorage(rdsProxyClient, request, progress);
+                    }
+                    return progress;
+                })
                 .then(progress -> ensureEngineSet(rdsProxyClient, progress))
                 .then(progress -> execOnce(progress, () ->
                                 updateDbInstance(proxy, request, rdsProxyClient, progress),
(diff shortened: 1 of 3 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -247,6 +247,13 @@ public static ModifyDbInstanceRequest modifyDbInstanceRequest(
             builder.iops(desiredModel.getIops());
         }
 
+        // In order to disable an instance autoscaling one has to reset `MaxAllocatedStorage` property.
+        // `MaxAllocatedStorage` would be reset upon an update if it is set to the current `AllocatedStorage` value.
+        // https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_PIOPS.StorageTypes.html#USER_PIOPS.Autoscaling
+        if (previousModel != null && previousModel.getMaxAllocatedStorage() != null && desiredModel.getMaxAllocatedStorage() == null) {
+            builder.maxAllocatedStorage(getAllocatedStorage(previousModel));
```
moataz-mhmd, line 254: Could you elaborate why setting `maxAllocatedStorage` to previousModel `AllocatedStorage` not the desiredModel 
i.e:              `builder.maxAllocatedStorage(getAllocatedStorage(desiredModel)); `

osdrv: In fact, we need to pick up the current value, hence neither previous nor desired resource state is a trustful source of this information, as autoscaling might has already changed the current `AllocatedStorage` value. It needs to be picked up from an explicit describe call. I'm going to address it in a jiffy.

moataz-mhmd (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/UpdateHandler.java
```diff
@@ -90,8 +90,12 @@ public UpdateHandler(final HandlerConfig config) {
                     return progress;
                 })
                 .then(progress -> ensureEngineSet(rdsProxyClient, progress))
-                .then(progress -> execOnce(progress, () ->
-                                updateDbInstance(proxy, request, rdsProxyClient, progress),
+                .then(progress -> execOnce(progress, () -> {
```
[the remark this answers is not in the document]
osdrv: Fixed by moving it to a proper chain step and wrapping in a try-catch block. Thank you!

moataz-mhmd (approved):

sungkkim (approved):

PR #93, opened 2021-12-07
Handle RDS InternalFailure

osdrv:
This commit adds a new ErrorCode called `InternalFailure`. It is returned by RDS API in a case of a service event.

An additional change to DBInstance and DBInstance role stabilization logic: errors are handled by `Commons.handleException/3` instead of the custom local try-catch block because it fails with a stack trace and an internal error status if RDS API returns a 5xx response code.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -47,8 +47,6 @@ import software.amazon.awssdk.services.rds.model.ProvisionedIopsNotAvailableInAz
 import software.amazon.awssdk.services.rds.model.SnapshotQuotaExceededException;
 import software.amazon.awssdk.services.rds.model.StorageQuotaExceededException;
 import software.amazon.awssdk.utils.StringUtils;
-import software.amazon.cloudformation.exceptions.CfnNotFoundException;
-import software.amazon.cloudformation.exceptions.CfnNotStabilizedException;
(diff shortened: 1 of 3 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -255,6 +250,11 @@ public BaseHandlerStd(final HandlerConfig config) {
                 .backoffDelay(config.getBackoff())
                 .makeServiceCall(NOOP_CALL)
                 .stabilize((request, response, proxyInvocation, model, context) -> isDbInstanceStabilized(proxyInvocation, model))
+                .handleError((request, exception, proxyInvocation, resourceModel, context) -> Commons.handleException(
+                        ProgressEvent.progress(resourceModel, context),
+                        exception,
+                        UPDATE_ASSOCIATED_ROLES_ERROR_RULE_SET
```
moataz-mhmd, line 256: Is this should be `DB_INSTANCE_ROLE_FAILED_TO_STABILIZE` rule set?

moataz-mhmd (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
osdrv: Decided to keep it as is, as NotStabilized is a terminal error

sungkkim (approved):

PR #97, opened 2022-01-24
Log Request and Response by Implementing Common RequestLogger 

moataz-mhmd:
Adding common LoggerWrapper with main keys:
- Silently fail if any exception thrown into LoggerWrapper and avoid NullPointerException as possible to not impact any resource because of using loggerWrapper.
- Log Request and Response using RequestLogger.
- Log any throwable that can happen during handle request and re-throw it without updating the stacktrace.
- Filter sensitive data (i.e: username, passwords... )from log in DBInstance and DBCluster.
- All logs in Json format so we can use CloudWatch logs insights on them if needed. 
- Adding wrap method to RequestLogger to RDS client request, response, and any exception if happened


aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/LoggerWrapper.java
```diff
@@ -0,0 +1,73 @@
...
+        }
+        ExceptionUtils.rethrow(throwable);
+    }
+
+    public void log(String message, Object object) {
+        log(message, object, Collections.emptyList());
+    }
+
+    //Adding data at the end for better experience. To make first line in log message meaningful 
+    public void log(String message, Object object, Collection<String> excludeParameters) {
```
osdrv, line 51: I believe the "exclude parameters" is somewhat an open-by-default approach, meaning that we can make a mistake and leak some data that would rather be omitted (compare this to "closed by default", requiring one to be explicit what is **enabled** instead). I believe we can take it even further and aim for a "safe by default" approach. One way to achieve it would be to:
- define a `LogDecorator` interface and ensure our overloaded logger always passes the log data through it
- define a `ObfuscateLogDecorator` class, defining the range of attributes that need to be obfuscated or replaced.
- The list of decorators could be passed along upon a logger instantiation and hence won't be required when a regular `log()` is being invoked.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -102,6 +104,9 @@
 
     protected HandlerConfig config;
 
+    private final Collection<String> SENSITIVE_PARAMETERS_PARENTS = ImmutableSet.of("desiredResourceState", "previousResourceState", "resourceModel");
```
osdrv, line 107: I believe we should aim for a more generic approach tbh, a hardcoded structure could be a limiting factor for the solution applicability in the future.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -308,38 +322,17 @@ protected boolean isAssociatedRolesEqual(
...
-                            addTagsToResourceRequest(arn, tagsToAdd),
-                            rdsClientProxyClient.client()::addTagsToResource
-                    );
-                    return ProgressEvent.progress(resourceModel, context);
-                });
-         */
+    private void logRequest(final LoggerWrapper loggerWrapper, final ResourceHandlerRequest<ResourceModel> request) {
+        loggerWrapper.log("Request: ", request, SENSITIVE_PARAMETERS_PARENTS);
+        loggerWrapper.log("DesiredResourceState: ", request.getDesiredResourceState(), SENSITIVE_PARAMETERS);
+        loggerWrapper.log("PreviousResourceState: ", request.getPreviousResourceState(), SENSITIVE_PARAMETERS);
```
osdrv, line 336: I believe we could rather inline these methods, as the approach pushes one to add more custom object-specific logging functions, which could pollute the code.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -224,13 +229,23 @@ public BaseHandlerStd(final HandlerConfig config) {
...
+                    context != null ? context : new CallbackContext(),
+                    proxy.newProxy(RdsClientBuilder::getClient),
+                    proxy.newProxy(Ec2ClientBuilder::getClient),
+                    logger
+            );
+            logResponse(loggerWrapper, progressEvent);
+        } catch (Throwable throwable) {
+            loggerWrapper.logThrowableAndRethrow(throwable);
+        }
+        return progressEvent;
```
osdrv, line 248: Looks a bit too massive. What if we turn it into a proxy-function instead? It also could be an annotation, which makes it even more elegant.

osdrv (approved):

ammokhov (dismissed):

ammokhov (approved):

osdrv (approved):

PR #99, opened 2022-01-27
[DBInstance] Rework DeleteHandler to use EventChain

osdrv:
This PR provides an alternative implementation for DBInstance
DeleteHandler. In particular, it removes a manual DeleteDBInstance
invocation, delegating it to a proxyInvocation.

We are removing an explicit handling of `DbInstanceDeletingException` as
it is being handled by `DELETE_DB_INSTANCE_ERROR_RULE_SET` already.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
@@ -4,10 +4,8 @@ import java.util.Optional;
 
 import org.apache.commons.lang3.BooleanUtils;
 
-import software.amazon.awssdk.awscore.exception.AwsServiceException;
 import software.amazon.awssdk.services.ec2.Ec2Client;
 import software.amazon.awssdk.services.rds.RdsClient;
-import software.amazon.awssdk.services.rds.model.DeleteDbInstanceResponse;
 import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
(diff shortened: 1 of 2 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
```diff
@@ -80,11 +63,4 @@ public DeleteHandler(final HandlerConfig config) {
                 ))
                 .done((deleteRequest, deleteResponse, proxyInvocation, model, context) -> ProgressEvent.defaultSuccessHandler(null));
     }
-
-    private boolean isDbInstanceDeletingException(final Exception exception) {
-        if (exception instanceof AwsServiceException) {
-            return Optional.ofNullable(exception.getMessage()).orElse("").contains(DB_INSTANCE_IS_BEING_DELETED_ERR);
-        }
-        return false;
-    }
```
moataz-mhmd, line 89: I think we will need to make ErrorRuleSet use errorMessage substring comparing in decision making in the future.
RdsClient can throw same error code but with different message which may lead to different behavior in handler.

moataz-mhmd (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
osdrv: Good point, will submit a separate PR. Thanks for the hint!

ammokhov (approved):

PR #101, opened 2022-01-31
Fix DBCluster Infinite Stabilization and Timeout If Role Already Exist or Already Deleted

moataz-mhmd:
Return in Failure only in adding roles or removing  roles to DBCluster. Otherwise resource will return IN_PROGRESS status and never complete the loop. Cloudformation will keep calling it up to 250 time then throw stabilization tiomeout.

Steps to trigger the issue:

1. Create DBCluster manually. Add associated role A to it
2. Do stack import with same cluster and Try to add associated role A, B to it

### DIFF
--- a/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
+++ b/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
@@ -198,7 +198,7 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
                             ADD_ASSOC_ROLES_ERROR_RULE_SET
                     ))
                     .success();
-            if (!progressEvent.isSuccess()) {
+            if (progressEvent.isFailed()) {
                 return progressEvent;
             }
         }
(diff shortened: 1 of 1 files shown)


osdrv (approved):

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -198,7 +198,7 @@ protected boolean isDBClusterDeleted(
                             ADD_ASSOC_ROLES_ERROR_RULE_SET
                     ))
                     .success();
-            if (!progressEvent.isSuccess()) {
+            if (progressEvent.isFailed()) {
                 return progressEvent;
```
ammokhov, line 202: I dont think this will lead to the correct result. 

`!progressEvent.isSuccess()` means that event it either failed or in progress. If it's failed we need to return to propagate to the customer. In case it's still in progress but lambda timeout is up, we do need to propagate the intermediate result back to the orchestrator so it will reinvoke and continue stabilization

ammokhov (changes requested):

PR #102, opened 2022-01-31
Fix DBInstance Infinite Stabilization and Timeout If Role Already Exist or Already Deleted

moataz-mhmd:
Return in Failure only in adding roles or removing roles to DBInstance. Otherwise resource will return IN_PROGRESS status and never complete the loop. Cloud formation will keep calling it up to 250 time then throw stabilization timeout.

Steps to trigger the issue:

   - Create DBInstace manually. Add associated role A to it
   - Do stack import with same instance and Try to add associated role A, B to it

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -449,7 +449,7 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
                             UPDATE_ASSOCIATED_ROLES_ERROR_RULE_SET
                     ))
                     .success();
-            if (!progressEvent.isSuccess()) {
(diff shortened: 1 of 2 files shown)


osdrv (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -457,7 +457,7 @@ protected void removeOldTags(
                             UPDATE_ASSOCIATED_ROLES_ERROR_RULE_SET
                     ))
                     .success();
-            if (!progressEvent.isSuccess()) {
+            if (progressEvent.isFailed()) {
                 return progressEvent;
```
ammokhov, line 461: same as in the other pr. I dont think this is right, we need to return also if it's in_progress. we basically can ignore only success bc the rest we need to propagate back to cfn

ammokhov (changes requested):

PR #104, opened 2022-02-01
Handle Error: Invalid Client Token Id

osdrv:
This PR adds an explicit error handler for `InvalidClientTokenId` error code.

AWS RDS Service might eventually return this error code if the credential token is stale. There is no better way to handle this rather than re-try in a bit: the error seems to be transient.

This PR treats `InvalidClientTokenId` as an internal service error.

### DIFF
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
@@ -19,7 +19,8 @@ public final class Commons {
     public static final ErrorRuleSet DEFAULT_ERROR_RULE_SET = ErrorRuleSet.builder()
             .withErrorCodes(ErrorStatus.failWith(HandlerErrorCode.ServiceInternalError),
                     ErrorCode.ClientUnavailable,
-                    ErrorCode.InternalFailure)
+                    ErrorCode.InternalFailure,
+                    ErrorCode.InvalidClientTokenId)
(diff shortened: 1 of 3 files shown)


aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
```diff
@@ -19,7 +19,8 @@ private Commons() {
     public static final ErrorRuleSet DEFAULT_ERROR_RULE_SET = ErrorRuleSet.builder()
             .withErrorCodes(ErrorStatus.failWith(HandlerErrorCode.ServiceInternalError),
                     ErrorCode.ClientUnavailable,
-                    ErrorCode.InternalFailure)
+                    ErrorCode.InternalFailure,
+                    ErrorCode.InvalidClientTokenId)
```
moataz-mhmd, line 23: Are we sure that CloudFormation will use new client token in retry?

osdrv: No, but we are silent observers here: we can either retry or bail out immediately. The best we can do is to retry this request and see what happens. Delegating it to the customer with a "invalid token" would certainly cause a "wtf" reaction.

moataz-mhmd (approved):

ammokhov (commented):
not sure if it's needed

osdrv:
Closing as the existing handler code is able to retry on an auth error

PR #105, opened 2022-02-02
[DBCluster] Set Engine and MasterUserPassword conditionally

osdrv:
This PR changes the way `ModifyDBCluster` request attributes are being set. In particular, `Engine` and `MasterUserPassword` are being set conditionally.

`Engine` should only be set if the previous and the desired values differ and the stack is not undergoing a rollback. Otherwise RDS API will return an error "Cannot modify engine version without a healthy primary instance in DB cluster".

`MasterUserPassword` should only be set if the values differ.

### DIFF
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/test/AbstractTestBase.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/test/AbstractTestBase.java
@@ -1,5 +1,6 @@
 package software.amazon.rds.common.test;
 
+import java.security.SecureRandom;
 import java.util.UUID;
 import java.util.function.Consumer;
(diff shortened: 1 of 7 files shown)


ammokhov (commented):
I think you also need to make changes in schema and add them to `conditionalCreateOnlyProperties`

osdrv:
@ammokhov `EngineVersion` is marked as `ConditionalCreateOnlyProperty`, but `MasterUserPassword` is declared as `writeOnly`. Would you still suggest moving the latter around to a more special group?

moataz-mhmd (approved):

moataz-mhmd (dismissed):

dbbh (approved):

moataz-mhmd (approved):

PR #108, opened 2022-02-07
[DBInstance] Handle drifting update requests

osdrv:
This code change adds an alternative Update handler branch to facilitate a DBInstance resource drift resolution.

The change checks the request `Driftable` flag and performs a scatter-status check for the supporting objects. The check validates `in-sync` status for the following objects:
  - DBParameterGroup
  - OptionGroup
  - DBClusterParameterGroup (if applicable)

Upon a complete sync, the handler would check if the instance requires a restart and performs a `reboot-await`.

### DIFF
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
@@ -4,7 +4,6 @@ import java.util.function.Function;
 
 import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.cloudformation.proxy.HandlerErrorCode;
-import software.amazon.cloudformation.proxy.OperationStatus;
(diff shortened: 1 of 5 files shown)


ammokhov (approved):
looks good!

moataz-mhmd (approved):
Just putting into your radar that the current behavior is doing `validateDeletionPolicy` step before fixing drift by checking
 if ( instance is a member in DBCluster && it is rollback update && Deletion policy is snapshot) then it will throw termination exception
 Other than that, It looks good to me.

osdrv:
@moataz-mhmd many thanks for the finding. Indeed, I intentionally decided to skip this part as per Uluru contract, `Drifting` is a no-change update. The check validates if the deletion policy was set to snapshot but the `dbclusteridentifier` is non-empty. Tbh, I fail to see its direct relation to the drifting process. Hence, merging it as is for now.

PR #110, opened 2022-02-08
Handle remove DBCluster from GlobalCluster case

moataz-mhmd:
Accept empty global cluster in schema which translated to removing DBCluster from GlobalCluster


osdrv:
Implemented in https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/111

PR #127, opened 2022-03-02
[DBInstance] Fix unstable collection order

osdrv:
This PR addresses a problem of an unstable ordering in the request objects created by Translator and supplementary midleware. The essence of the problem is in the hashCode computed for a particular request by the CFN core library. A request hashCode is used as a part of the callback context call graph key and is expected to be invariant in between multiple invocations of the same call chain. If a collection order is unstable, it can change the hash entrophy and result to a
distinct hashCode sum. This causes CFN to loose track of the previous progress and restart it from the beginning.

On practice it means that an unstable request object hash code could cause CFN to re-run an already executed stage twice, causing the service to terminate the stack with an error.


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -482,17 +483,17 @@ public static String translateDBParameterGroupFromSdk(final software.amazon.awss
...
     }
 
-    public static List<software.amazon.awssdk.services.rds.model.Tag> translateTagsToSdk(final Collection<Tag> tags) {
+    public static Set<software.amazon.awssdk.services.rds.model.Tag> translateTagsToSdk(final Collection<Tag> tags) {
         return streamOfOrEmpty(tags)
                 .map(tag -> software.amazon.awssdk.services.rds.model.Tag.builder()
                         .key(tag.getKey())
                         .value(tag.getValue())
                         .build()
                 )
```
moataz-mhmd, line 495: Should we sort here to ensure same order between all handler responses?
```
.sorted((t1, t2)->o1.getKey().compareTo(o2.getKey()))
```

moataz-mhmd (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
osdrv: I believe we would rather eliminate the order change, i.e. keep it "as given"

dbbh (approved):

PR #135, opened 2022-03-11
[DBInstance] Fix Endpoint.Port datatype from Integer to String

osdrv:
This commit changes the data type definition for `Endpoint.Port`. The new implementation happened to copy the type definition from AWS RDS SDK, which declares this attribute as an Integer.

Having this fix in place will unblock customers using `Endpoint.Port` attribute in string functions.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
@@ -434,7 +434,7 @@ public class Translator {
     private static Endpoint translateEndpointFromSdk(software.amazon.awssdk.services.rds.model.Endpoint endpoint) {
         return Endpoint.builder()
                 .address(endpoint.address())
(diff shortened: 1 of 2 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -434,7 +434,7 @@ public static DescribeDbEngineVersionsRequest describeDbEngineVersionsRequest(
     private static Endpoint translateEndpointFromSdk(software.amazon.awssdk.services.rds.model.Endpoint endpoint) {
         return Endpoint.builder()
                 .address(endpoint.address())
-                .port(endpoint.port())
+                .port(endpoint.port() == null ? null : endpoint.port().toString())
```
moataz-mhmd, line 437: No sure if this statement is useful if port is string in schema.

dbbh (approved):

PR #172, opened 2022-04-26
[Common] Introduce `IdentifierFactory`

osdrv:
This commit introduces a new util factory class called `IdentifierFactory`. The aim is to provide a generic and safe approach
for new identifier generation.

The current approach, which is adopted across all RDS CFN resources, uses cloudofrmation `IdentifierUtils` utility, which appears to be missing the double hyphens case and performs no null-safety checks on identifier components: stack id, logical resource identifier and request token.

The newly added util class generalizes the identifier generation approach and hides null-safety checks.

This commit affects all RDS CFN resources. Some areas of the code were changed due to an auto-formatting and auto-rearranges.

### DIFF
--- a/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
+++ b/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
@@ -7,13 +7,19 @@ import software.amazon.cloudformation.proxy.Logger;
 import software.amazon.cloudformation.proxy.ProgressEvent;
 import software.amazon.cloudformation.proxy.ProxyClient;
 import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
-import software.amazon.cloudformation.resource.IdentifierUtils;
 import software.amazon.rds.common.handler.Commons;
 import software.amazon.rds.common.handler.HandlerConfig;
 import software.amazon.rds.common.handler.Tagging;
+import software.amazon.rds.common.util.IdentifierFactory;
 
 
 public class CreateHandler extends BaseHandlerStd {
+
+    private final static IdentifierFactory groupIdentifierFactory = new IdentifierFactory(
+            STACK_NAME,
+            RESOURCE_IDENTIFIER,
+            MAX_LENGTH_GROUP_NAME
+    );
 
     public CreateHandler() {
         this(HandlerConfig.builder().build());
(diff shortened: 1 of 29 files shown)


aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
```diff
@@ -49,7 +55,7 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(final Amaz
     private ProgressEvent<ResourceModel, CallbackContext> createDbClusterParameterGroup(final AmazonWebServicesClientProxy proxy,
                                                                                         final ProxyClient<RdsClient> proxyClient,
                                                                                         final ProgressEvent<ResourceModel, CallbackContext> progress,
-                                                                                        final Tagging.TagSet tags ) {
+                                                                                        final Tagging.TagSet tags) {
```
Outige, line 58: Very nice change 😃

osdrv: This is likely caused by auto-formatter: I run 3 things upon every file change:
- optimize imports
- reformat code
- rearrange code
This change is quite full of these whitespace fixes due to this automation.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
```diff
@@ -43,12 +50,11 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
...
                 !isDBClusterMember(resourceModel)) {
-            snapshotIdentifier = generateResourceIdentifier(
-                    Optional.ofNullable(request.getStackId()).orElse(STACK_NAME),
-                    SNAPSHOT_PREFIX + Optional.ofNullable(request.getLogicalResourceIdentifier()).orElse(RESOURCE_IDENTIFIER),
-                    request.getClientRequestToken(),
-                    SNAPSHOT_MAX_LENGTH
-            );
+            snapshotIdentifier = snapshotIdentifierFactory.newIdentifier()
+                    .withStackId(request.getStackId())
+                    .withResourceId(SNAPSHOT_PREFIX + Optional.ofNullable(request.getLogicalResourceIdentifier()).orElse(RESOURCE_IDENTIFIER))
```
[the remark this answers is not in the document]
osdrv: If `request.getLogicalResourceIdentifier()` return null, string concatenation fill throw an exception. `Optional` is guarding against this case. `defaultResourceId` is fine, but effectively won't be used here given the non-empty prefix (`SNAPSHOT_PREFIX`) is provided.

Outige, line 55: nit:
We have set `defaultResourceId`. But we won't be able ever use the default value because we call `.withResourceId(SNAPSHOT_PREFIX + ....` and the `defaultResourceId` won't be used because we will never pass an empty resourceId.

Do you think we could replace this line with something along the lines of:
`.withResourceId(request.getLogicalResourceIdentifier() == null ? null : SNAPSHOT_PREFIX + request.getLogicalResourceIdentifier())`
This way we can actually pass a null value to trigger the use of our default value?

osdrv: Replaced with a `StringUtils.prependIfMissing(request.getLogicalResourceIdentifier(), SNAPSHOT_PREFIX)`. I aimed to avoid the ternary operator due to its ultimate clumsiness, `StringUtils` came up to the rescue as `prependIfMissing` is null-safe: it simply returns null if the first argument is also null. We will have an implicit fallback in this case.

dbbh (approved):

Outige (approved):

PR #178, opened 2022-05-06
[DBClusterParameterGroup] Handle Create Call while Stack Having IAM Role with Tag Condition

moataz-mhmd:
*Description of changes:*
When customer specify stack IAM role with tag conditions. Create call will fail with `AccessDenied` exception because it contains only system tags and apply other tags by `AddTagsToResource`.

I.e: Customer created this stack with stack level tag `foo -> bar`
```
AWSTemplateFormatVersion: 2010-09-09
Description: RDS DBClusterParameterGroup Example Stack

Resources:
  RDSDBClusterParameterGroup:
    Type: AWS::RDS::DBClusterParameterGroup
    DeletionPolicy: Delete
    Properties:
      Parameters:
        character_set_database: utf32
      Family: aurora5.6
      Description: "Any Description"
```

And assigned IAM Role to execute this stack
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": "rds:*",
            "Resource": "*",
            "Condition": {
                "StringEquals": {
                    "rds:req-tag/foo": "bar"
                }
            }
        }
    ]
}
```
That means `CreateDBClusterParameterGroup` call need to have `--tags foo -> bar` Or the whole call will fail with `AccessDenied` Exception. 

**Approach**
Change create to include all possible tags (Stack, System, and Resource) together.
If failed with access denied that means there is possibility to apply soft fail for stack level tags, We trigger create call again with system tags only to ensure they are attached as they do not need permission. Then code flow will follow regular soft fail procedure for stack level tags.
 
### DIFF
--- a/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
+++ b/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
@@ -1,8 +1,11 @@
 package software.amazon.rds.dbclusterparametergroup;
 
+import java.util.Collections;
+
 import com.amazonaws.util.StringUtils;
 import software.amazon.awssdk.services.rds.RdsClient;
 import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
+import software.amazon.cloudformation.proxy.HandlerErrorCode;
 import software.amazon.cloudformation.proxy.Logger;
 import software.amazon.cloudformation.proxy.ProgressEvent;
 import software.amazon.cloudformation.proxy.ProxyClient;
@@ -35,19 +38,15 @@ public class CreateHandler extends BaseHandlerStd {
                                                                           final CallbackContext callbackContext,
                                                                           final ProxyClient<RdsClient> proxyClient,
                                                                           final Logger logger) {
-        final Tagging.TagSet systemTags = Tagging.TagSet.builder()
+        final Tagging.TagSet allTags = Tagging.TagSet.builder()
                 .systemTags(Tagging.translateTagsToSdk(request.getSystemTags()))
-                .build();
-
-        final Tagging.TagSet extraTags = Tagging.TagSet.builder()
                 .stackTags(Tagging.translateTagsToSdk(request.getDesiredResourceTags()))
                 .resourceTags(Translator.translateTagsToSdk(request.getDesiredResourceState().getTags()))
                 .build();
 
         return ProgressEvent.progress(request.getDesiredResourceState(), callbackContext)
                 .then(progress -> setDbClusterParameterGroupNameIfMissing(request, progress))
-                .then(progress -> createDbClusterParameterGroup(proxy, proxyClient, progress, systemTags))
-                .then(progress -> updateTags(proxy, proxyClient, progress, Tagging.TagSet.emptySet(), extraTags))
+                .then(progress -> createDbClusterPGWithTags(proxy, proxyClient, progress, allTags))
                 .then(progress -> applyParameters(proxy, proxyClient, progress.getResourceModel(), progress.getCallbackContext()))
                 .then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
     }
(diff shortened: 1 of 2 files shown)


aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
```diff
@@ -35,18 +36,19 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(final Amaz
...
 
         final Tagging.TagSet extraTags = Tagging.TagSet.builder()
                 .stackTags(Tagging.translateTagsToSdk(request.getDesiredResourceTags()))
                 .resourceTags(Translator.translateTagsToSdk(request.getDesiredResourceState().getTags()))
                 .build();
 
+        final Tagging.TagSet allTags = extraTags.toBuilder()
+                .systemTags(Tagging.translateTagsToSdk(request.getSystemTags()))
+                .build();
+
```
osdrv, line 48: nit: I would suggest flipping it the other way around: all tags is the primary declaration and the extra tags as a subset.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
```diff
@@ -35,18 +36,19 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(final Amaz
...
         return ProgressEvent.progress(request.getDesiredResourceState(), callbackContext)
                 .then(progress -> setDbClusterParameterGroupNameIfMissing(request, progress))
-                .then(progress -> createDbClusterParameterGroup(proxy, proxyClient, progress, systemTags))
+                .then(progress -> createDbClusterPGWithAllTagsOrSystemTags(proxy, proxyClient, progress, allTags))
```
osdrv, line 51: nit: too much implementation details in the method name. Consider: `safeCreateDBClusterParameterGroup`

Outige (approved):
LGTM

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
```diff
@@ -73,6 +75,18 @@ private ProgressEvent<ResourceModel, CallbackContext> createDbClusterParameterGr
                 });
     }
 
+    private ProgressEvent<ResourceModel, CallbackContext> createDbClusterPGWithAllTagsOrSystemTags(final AmazonWebServicesClientProxy proxy,
+                                                                                        final ProxyClient<RdsClient> proxyClient,
+                                                                                        final ProgressEvent<ResourceModel, CallbackContext> progress,
+                                                                                        final Tagging.TagSet allTags) {
+        ProgressEvent<ResourceModel, CallbackContext> progressEvent = createDbClusterParameterGroup(proxy, proxyClient, progress, allTags);
+        if(HandlerErrorCode.AccessDenied.equals(progressEvent.getErrorCode())) { //Resource is subject to soft fail on stack level tags.
```
osdrv, line 83: nit: code formatting

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
```diff
@@ -73,6 +75,18 @@ private ProgressEvent<ResourceModel, CallbackContext> createDbClusterParameterGr
...
+            Tagging.TagSet systemTags = Tagging.TagSet.builder().systemTags(allTags.getSystemTags()).build();
+            return createDbClusterParameterGroup(proxy, proxyClient, progress, systemTags);
```
osdrv, line 85: nit: Here is an idea: `createDbClusterParameterGroup` might set a flag like: `taggingOk: true`, which would be checked by `updateTags` and skip this step if it was an optimistic tagging case.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
moataz-mhmd: done!

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
moataz-mhmd: Done! I updated method logic and naming. Please check if you have better naming.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CreateHandler.java
moataz-mhmd: Done!

osdrv (approved):

Outige (approved):

PR #187, opened 2022-05-17
[DBSubnetGroup] Accept duplicate SubnetIds

moataz-mhmd:
*Description of changes:*
This commit changes DBSubnetGroup schema to tolerate duplicate `SubnetIds` and drop those duplicate in client request in order to not break current stacks with duplicate `SubnetIds`.

### DIFF
--- a/aws-rds-dbsubnetgroup/aws-rds-dbsubnetgroup.json
+++ b/aws-rds-dbsubnetgroup/aws-rds-dbsubnetgroup.json
@@ -12,7 +12,7 @@
     },
     "SubnetIds": {
       "type": "array",
-      "uniqueItems": true,
+      "uniqueItems": false,
       "items": {
         "type": "string"
       }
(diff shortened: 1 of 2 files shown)


aws-rds-dbsubnetgroup/aws-rds-dbsubnetgroup.json
```diff
@@ -12,7 +12,7 @@
     },
     "SubnetIds": {
       "type": "array",
-      "uniqueItems": true,
+      "uniqueItems": false,
```
Outige, line 15: Why do we need to update this if we change this? The customers input SubnetIDs are already converted into a set which should enforce uniqueness.

moataz-mhmd: Because of backward compatibility. Current stacks have duplicate `SubnetIds` and working fine. If we did not change it, We would break already working stacks.

Outige (dismissed):

Outige (approved):

PR #190, opened 2022-05-19
[DBInstance] Avoid unnecessary rename upon a DBInstance update

osdrv:
This commit eliminates an unnecessary rename + reboot upon a database instance update. The reason it happens is because `Translator.modifyDBInstanceRequest` sets `ModifyDBInstanceRequest.newInstanceIdentifier` unconditionally, causing RDS to interpret this as an attempt to rename the instance. This automatically causes a reboot. By setting this attribute conditionally we eliminate the reboot and avoid a potential downtime.


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -285,7 +287,7 @@ public static ModifyDbInstanceRequest modifyDbInstanceRequest(
                 .caCertificateIdentifier(desiredModel.getCACertificateIdentifier())
                 .copyTagsToSnapshot(desiredModel.getCopyTagsToSnapshot())
                 .dbInstanceClass(desiredModel.getDBInstanceClass())
-                .dbInstanceIdentifier(desiredModel.getDBInstanceIdentifier())
+                .dbInstanceIdentifier(currentIdentifier)
```
moataz-mhmd, line 290: I wonder how customers can rename DbInstance. I do not see `NewDBInstanceIdentifier` attribute in schema. Then the only way is by replacement after updating `DBInstanceIdentifier` then CFN will detect it is create only then delete and create new instance with same attribute. Is this behavior on intention or I am missing something.

moataz-mhmd (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
osdrv: That's a good point. I'll get it fixed. TY!

Outige (dismissed):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
moataz-mhmd: Just heads up there is a loop can happen. DescribeDBInstance will return new `DBInstanceIdentifier` which will be detected as drift from source stack, Then CloudFormation will replace again to achieve old `DBInstanceIdentifier` but modify will apply `NewDBInstanceIdentifier` again and it generate loop.

osdrv: Noted and fixed.

moataz-mhmd (approved):

khebul (approved):

PR #191, opened 2022-05-23
[DBInstance] Restore original `dbInstanceIdentifier` case spelling

osdrv:
This commit alters ReadHandler behavior in a way it interprets a response from RDS API. RDS and CFN primary identifier contract differs significantly: whereas CFN treats primary identifiers as unconditionally immutable, RDS operates in a case-insensitive mode and coalesces all primary identifiers to lowercase. This contract-level disagreement causes an unpredictable behavior and triggers a drift detector immediately.

This commit instructs the read handler to restore the original `dbInstanceIdentifier` from `ResourceModel` in the resulting `ProgressEvent`. Having this in place allows RDS to keep operating in case-insensitive mode and return the original identifier matching the customer template value back to CFN.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -399,6 +399,39 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
         return StringUtils.isNotBlank(model.getDBClusterIdentifier());
     }
 
+    protected ResourceModel restoreIdentifier(final ResourceModel observed, final ResourceModel original) {
+        if (StringUtils.isBlank(original.getDBInstanceIdentifier()) ||
+            StringUtils.equals(observed.getDBInstanceIdentifier(), original.getDBInstanceIdentifier())) {
+            return observed;
+        }
+        observed.setDBInstanceIdentifier(original.getDBInstanceIdentifier());
+        return observed;
+    }
+
+    protected DBInstance restoreIdentifier(final DBInstance dbInstance, final ResourceModel model) {
+        if (StringUtils.isBlank(model.getDBInstanceIdentifier()) ||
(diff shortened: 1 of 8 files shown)


khebul (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -399,6 +399,39 @@ protected boolean isDBClusterMember(final ResourceModel model) {
         return StringUtils.isNotBlank(model.getDBClusterIdentifier());
     }
 
+    protected ResourceModel restoreIdentifier(final ResourceModel observed, final ResourceModel original) {
```
moataz-mhmd, line 402: Why we return ResourceModel here, Should we make this method `void` as it is editing value in observed object only and always returning observed?

moataz-mhmd (approved):
Should we make it safer by checking string `equalsIgnoreCase` then return actual in case of false. 

Just to put into your radar that customer can rename DBInstance or DBCluster out of band during handler execution. This case will fail anyway in next call with does not exist exception.  

ref: https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#equalsIgnoreCase(java.lang.String)

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
osdrv: I decided to stick to the functional interface as the new CFN framework is somewhat promoting it all around. In this case it simplifies passing the arguments around as in ReadHandler:
```java
return ProgressEvent.success(restoreIdentifier(Translator.translateDbInstanceFromSdk(dbInstance), request.getDesiredResourceState()), context);
```
You are right: this is not a real functional approach as it comes with a side effect. Given that all use cases of these methods converge to passing a brand new object that was transformed from the response, I took this shortcut. In other words, I'm not concerned about having this side effect atm, but if I had to be 100% pedantic about it, I'd stick to a functional approach with a modified copy of the object returned back. Hope it makes sense.

PR #194, opened 2022-05-26
[OptionGroup] Implement Soft Fail on stack tags

Outige:
**Description of changes**:

Soft fails means ignoring permission denied exception and proceed with resource logic.This PR implements soft fail in OptionGroup:
- Hard fail on add/remove resource tags
- Soft fail on add/remove system tags
- Soft fail on add/remove stack level tags

Signed-off-by: Tieg O'Sullivan [tiegaws@amazon.com](mailto:tiegaws@amazon.com)

### DIFF
--- a/aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/BaseHandlerStd.java
+++ b/aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/BaseHandlerStd.java
@@ -25,9 +25,12 @@ import software.amazon.rds.common.error.ErrorRuleSet;
 import software.amazon.rds.common.error.ErrorStatus;
 import software.amazon.rds.common.handler.Commons;
 import software.amazon.rds.common.handler.HandlerConfig;
+import software.amazon.rds.common.handler.Tagging;
 import software.amazon.rds.common.logging.LoggingProxyClient;
 import software.amazon.rds.common.logging.RequestLogger;
 import software.amazon.rds.common.printer.FilteredJsonPrinter;
+
+import software.amazon.rds.common.handler.Tagging;
 
 public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
 
@@ -109,8 +112,8 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
             final AmazonWebServicesClientProxy proxy,
             final ProxyClient<RdsClient> proxyClient,
             final ProgressEvent<ResourceModel, CallbackContext> progress,
-            final Map<String, String> previousTags,
-            final Map<String, String> desiredTags
+            final Tagging.TagSet previousTags,
+            final Tagging.TagSet desiredTags
     ) {
         return proxy.initiate("rds::tag-option-group", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
                 .translateToServiceRequest(Translator::describeOptionGroupsRequest)
@@ -123,21 +126,25 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
                         DEFAULT_OPTION_GROUP_ERROR_RULE_SET
                 ))
                 .done((describeRequest, describeResponse, invocation, resourceModel, ctx) -> {
+                    final Tagging.TagSet tagsToAdd = Tagging.exclude(desiredTags, previousTags);
+                    final Tagging.TagSet tagsToRemove = Tagging.exclude(previousTags, desiredTags);
+
+                    if (tagsToRemove.isEmpty() && tagsToAdd.isEmpty()) {
+                        return progress;
+                    }
+
                     final String arn = describeResponse.optionGroupsList().stream().findFirst().get().optionGroupArn();
-
-                    final Set<Tag> previousTagSet = Translator.translateTagsToModelResource(previousTags);
-                    final Set<Tag> desiredTagSet = Translator.translateTagsToModelResource(desiredTags);
-                    final Set<Tag> tagsToRemove = Sets.difference(previousTagSet, desiredTagSet);
-                    final Set<Tag> tagsToAdd = Sets.difference(desiredTagSet, previousTagSet);
-
-                    proxyClient.injectCredentialsAndInvokeV2(
-                            Translator.removeTagsFromResourceRequest(arn, tagsToRemove),
-                            proxyClient.client()::removeTagsFromResource
-                    );
-                    proxyClient.injectCredentialsAndInvokeV2(
-                            Translator.addTagsToResourceRequest(arn, tagsToAdd),
-                            proxyClient.client()::addTagsToResource
-                    );
+                    try {
+                        Tagging.removeTags(proxyClient, arn, Tagging.translateTagsToSdk(tagsToRemove));
+                        Tagging.addTags(proxyClient, arn, Tagging.translateTagsToSdk(tagsToAdd));
+                    } catch (Exception exception) {
+                        return Commons.handleException(
+                                progress,
+                                exception,
+                                Tagging.bestEffortErrorRuleSet(tagsToAdd, tagsToRemove, Tagging.SOFT_FAIL_IN_PROGRESS_TAGGING_ERROR_RULE_SET, Tagging.HARD_FAIL_TAG_ERROR_RULE_SET)
+                                        .orElse(DEFAULT_OPTION_GROUP_ERROR_RULE_SET)
+                        );
+                    }
                     return ProgressEvent.progress(resourceModel, ctx);
                 });
     }
--- a/aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CallbackContext.java
+++ b/aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CallbackContext.java
@@ -7,4 +7,5 @@ import software.amazon.cloudformation.proxy.StdCallbackContext;
 @lombok.ToString
 @lombok.EqualsAndHashCode(callSuper = true)
 public class CallbackContext extends StdCallbackContext {
+    private String optionGroupArn;
 }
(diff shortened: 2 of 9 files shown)


aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/Translator.java
```diff
@@ -160,7 +156,8 @@ static Set<Tag> translateTagsToModelResource(final Map<String, String> tags) {
                         .key(entry.getKey())
                         .value(entry.getValue())
                         .build())
-                .collect(Collectors.toSet());
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+//                .collect(Collectors.toSet());
```
Outige, line 160: Needs to go

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CreateHandler.java
```diff
@@ -49,31 +61,39 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
...
+                .then(progress -> updateTags(proxy, proxyClient, progress, Tagging.TagSet.emptySet(), extraTags))
                 .then(progress -> {
                     if (CollectionUtils.isNullOrEmpty(progress.getResourceModel().getOptionConfigurations())) {
                         return progress;
                     }
                     return updateOptionGroup(proxy, proxyClient, progress);
                 })
+                .then(progress -> {
+                    return progress;
+                })
```
osdrv, line 74: No-op step, safe to remove

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CreateHandler.java
```diff
@@ -49,31 +61,39 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
...
                 .then(progress -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
     }
+
+    private ProgressEvent<ResourceModel, CallbackContext> createDbClusterParameterGroup(final AmazonWebServicesClientProxy proxy,
+                                                                                        final ProxyClient<RdsClient> proxyClient,
+                                                                                        final ProgressEvent<ResourceModel, CallbackContext> progress,
+                                                                                        final Tagging.TagSet tags) {
+        return proxy.initiate("rds::create-option-group", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
```
osdrv, line 82: I believe it needs to be handled in a 2-tries way: 1st with all the tags available and the second with system tags only. Please see this ref as an example: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/CreateHandler.java#L58-L71

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/ReadHandler.java
```diff
@@ -47,6 +49,7 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
                     final OptionGroup optionGroup = describeResponse.optionGroupsList().stream().findFirst().get();
                     final List<OptionConfiguration> optionConfigurations = Translator.translateOptionConfigurationsFromSdk(optionGroup.options());
                     final List<Tag> tags = listTags(proxyInvocation, optionGroup.optionGroupArn());
+                    context.setOptionGroupArn(optionGroup.optionGroupArn());
```
osdrv, line 52: As it is a part of ReadHandler, which is only invoked as a chain finalizer, I assume this arn set is a no-op. Could you please check it?

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/Translator.java
```diff
@@ -1,13 +1,8 @@
 package software.amazon.rds.optiongroup;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
osdrv, line 3: Please don't. `util` is significantly bigger than the range of the classes we're importing here.

aws-rds-optiongroup/src/test/java/software/amazon/rds/optiongroup/CreateHandlerTest.java
```diff
@@ -9,26 +10,35 @@
...
 import java.time.Duration;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import jdk.nashorn.internal.runtime.regexp.joni.exception.InternalException;
```
osdrv, line 21: Omg, what's that? I think it was an import glitch.

aws-rds-optiongroup/src/test/java/software/amazon/rds/optiongroup/CreateHandlerTest.java
```diff
@@ -9,26 +10,35 @@
...
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 import lombok.Getter;
+import software.amazon.awssdk.awscore.exception.AwsErrorDetails;
 import software.amazon.awssdk.services.rds.RdsClient;
+import software.amazon.awssdk.services.rds.model.*;
```
osdrv, line 34: Please disambiguate this asterisk

aws-rds-optiongroup/src/test/java/software/amazon/rds/optiongroup/CreateHandlerTest.java
```diff
@@ -9,26 +10,35 @@
...
+import software.amazon.cloudformation.proxy.*;
```
osdrv, line 35: Same here: let's disambiguate it.

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CreateHandler.java
Outige: If I understand correctly, the 2-tries way would be to address the IAM role with tags issue? As addressed in this PR: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/188.

Do you think I should do this in this PR or 2 separate PRs?

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/ReadHandler.java
Outige: I think I need some clarification offline about what exactly you are asking here.

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CreateHandler.java
moataz-mhmd: When I implemented soft failing before, We did not know about IAM issue. THat's why we have 2 PRs (soft fail, safe create). Now, We know this approach is not valid any more without safe create that Oleg mentioned. We can not merge code that we sure it is not covering all cases(having a bug).

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CallbackContext.java
```diff
@@ -7,4 +7,5 @@
 @lombok.ToString
 @lombok.EqualsAndHashCode(callSuper = true)
 public class CallbackContext extends StdCallbackContext {
+    private String optionGroupArn;
```
moataz-mhmd, line 10: Could you elaborate why this needed. Why you are using it by calling `context.getOptionGroupArn()`?

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/CreateHandler.java
Outige: I've added the IAM softfail now

moataz-mhmd (dismissed):

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/BaseHandlerStd.java
```diff
@@ -87,10 +87,13 @@ protected abstract ProgressEvent<ResourceModel, CallbackContext> handleRequest(
             final ProxyClient<RdsClient> proxyClient,
             final Logger logger);
 
-    protected ProgressEvent<ResourceModel, CallbackContext> updateOptionGroup(
+    protected ProgressEvent<ResourceModel, CallbackContext> updateOptionGroupIfRequired(
```
osdrv, line 90: I personally am not a big fan of `callIfCondition` methods. Another alternative to that would be:
`if condition { call(); }`. The latter option is semantically cleaner. The condition could be simplified with a supplementary method, like: `isShouldCall()`.

aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/BaseHandlerStd.java
```diff
@@ -87,10 +87,13 @@ protected abstract ProgressEvent<ResourceModel, CallbackContext> handleRequest(
...
             final AmazonWebServicesClientProxy proxy,
             final ProxyClient<RdsClient> proxyClient,
             final ProgressEvent<ResourceModel, CallbackContext> progress) {
+        if (CollectionUtils.isNullOrEmpty(progress.getResourceModel().getOptionConfigurations())) {
+            return progress;
+        }
```
osdrv, line 96: And this could be moved to `shouldUpdateOptionGroup` method.

osdrv (approved):

moataz-mhmd (approved):

khebul (approved):

PR #201, opened 2022-06-07
[EventSubscription] Add EventSubscriptionName as CreateOnly parameter

mtz-az:
*Description of changes:*
EventSubscriptionName is immutable and replacement is needed in case of changing it.


osdrv:
From the functional perspective, this change is a no-op as `SubscriptionName` is already defined as the primary identifier, i.e. CFN will attempt to replace it anyways. Is it related to the documentation or something?

moataz-mhmd (commented):
Yes, Documentation is generated from schema. If we did not add it, It will appear as mutable parameter. Also CFN team mentioned they do not manage this case of primary identifier is always immutable. You need to specify it as `CreateOnly` to trigger the replacement.

Outige (approved):

osdrv (approved):

PR #206, opened 2022-06-10
[DBInstance] Fix repeating `Create*` call failures on tag softfail

osdrv:
This commit fixes an issue in `CreateHandler`. When the handler soft-fails a `Create*` call containing all tags, it attempts to invoke the same method with system tags only. If this call succeeds, it enters a stabilization cycle. A `ProgressEvent` is re-entrant, meaning that any code that is not guarded by the special `ProgressEvent` sentinels would be re-executed every time a stabilizer returns an in-progress signal.

This effect had an impact on the `safeAddTags` method: whenever there was a downgrade to system-tags only, the original all-tags method was still invoked every time the system-tags only stabilizer returned an in-progress. This commit ensures that the all-tags `Create*` attempt will only fail once and won't be triggered upon a re-entrance.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
@@ -134,15 +134,28 @@ public class CreateHandler extends BaseHandlerStd {
             final HandlerMethod<ResourceModel, CallbackContext> handlerMethod,
             final Tagging.TagSet allTags
     ) {
-        return (proxy, rdsProxyClient, progress, tagSet) -> {
-            final ProgressEvent<ResourceModel, CallbackContext> result = handlerMethod.invoke(proxy, rdsProxyClient, progress, allTags);
-            if (HandlerErrorCode.AccessDenied.equals(result.getErrorCode())) {
-                final Tagging.TagSet systemTags = Tagging.TagSet.builder().systemTags(allTags.getSystemTags()).build();
-                return handlerMethod.invoke(proxy, rdsProxyClient, progress, systemTags);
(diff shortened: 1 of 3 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -134,15 +134,28 @@ private HandlerMethod<ResourceModel, CallbackContext> safeAddTags(
...
-            final ProgressEvent<ResourceModel, CallbackContext> result = handlerMethod.invoke(proxy, rdsProxyClient, progress, allTags);
-            if (HandlerErrorCode.AccessDenied.equals(result.getErrorCode())) {
-                final Tagging.TagSet systemTags = Tagging.TagSet.builder().systemTags(allTags.getSystemTags()).build();
-                return handlerMethod.invoke(proxy, rdsProxyClient, progress, systemTags);
+        return (proxy, rdsProxyClient, progress, tagSet) -> progress.then(p -> {
+            final CallbackContext context = p.getCallbackContext();
+            if (context.isSoftFailTags()) {
+                return p;
+            }
+            final ProgressEvent<ResourceModel, CallbackContext> result = handlerMethod.invoke(proxy, rdsProxyClient, p, allTags);
```
Outige, line 142: nit: Should we name this result something more meaningful?

Outige (approved):

khebul (commented):
DBCluster resource uses exactly the same code for soft-fail tagging. Could you replicate the same fix there?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
osdrv: I believe the scope of this lambda is very compact, hence the naming

PR #207, opened 2022-06-14
[Common] Implement `Tagging.safeCreate`

osdrv:
This commit implements a method `Tagging.safeCreate` that is supposed to handle tag soft-failing logic. Effectively, this method was extracted from `DBInstance.CreateHandler` implementation. The method accepts a regular set of attributes like: proxy, progress event etc. It also accepts an instance of `HandlerMethod` and a `Tagging.TagSet`. The
method then will attempt to invoke the `HandlerMethod` instance. If this invocation results in an error, the `safeCreate` method will inspect the error code. A `HandlerErrorCode.AccessDenied` would be treated as a missing `rds:AddTagsToResource` permissions and the method will retry the create attempt, but with system tags only. This would be the second invocation of `HandlerMethod`. The result of the second invocation is then returned as is.

This commit introduces a mixin data class called `TaggingContext` under the commons package. The purpose of this class is to be added to a `CallbackContext` class and exposing its availability by an explicit `implements TaggingContext.Provider` class annotation. Any further tagging-related flags are supposed to be located in this mixin class.


aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
```diff
@@ -200,6 +201,38 @@ public static ErrorRuleSet bestEffortErrorRuleSet(
...
+            }
+            final ProgressEvent<M, C> allTagsResult = handlerMethod.invoke(proxy, rdsProxyClient, p, allTags);
+            if (allTagsResult.isFailed()) {
+                if (HandlerErrorCode.AccessDenied.equals(allTagsResult.getErrorCode())) {
+                    context.getTaggingContext().setSoftFailTags(true);
+                    return ProgressEvent.progress(allTagsResult.getResourceModel(), context);
+                }
+                return allTagsResult;
+            }
+            allTagsResult.getCallbackContext().getTaggingContext().setAddTagsComplete(true);
```
Outige, line 224: Should we not have a check at the top of this logic to see if we've successfully added the tags?

Something like:
```
            if (context.getTaggingContext().isSoftFailTags() || context.getTaggingContext().isAddTagsComplete()) {
                return p;
            }
```

I see on line 108 of `CreateHandler` isAddTagsComplete() is being called. But I can't quite make out what is happening there.

If line 108 is addressing this issue. Then the problem becomes a nit. Can we not rather just move the logic to the code I suggested in this comment?

osdrv: I think the implicit part in this case is that Uluru will check if it has executed this chain already. What we are trying to solve here is: Uluru will only memorize the chain if it succeeded, but not when it failed. This is the essence of the fix we merged recently: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/206. I can rephrase this logic as:
```
part 1
    if failed to create with all tags before (`isSoftFailTags` is set)
        then skip to the second part immediately
    try to create a resource with all tags
    if it fails
        if it is an access denied
            set `isSoftFailTags` -> true (so the next invocation of this chain will skip this part right away)
            return an in-progress event so it could be handled by part 2
        else
            return failed event
    else
        set `isAddTagsComplete` -> true (so this flag can be used by the downstream consumers of this event to distinguish between a successful create with all tags and a create with system tags only)
        return an in-progress event so it could be handled by part 2
part2
    if not `isSoftFailTags` (i.e. the first part never failed)
        return immediately
    create with system tags only
    return this result unconditionally
```

Uluru event chains are re-entrant. I.e. executed once, if it returns an in-progress event, it will be re-invoked again by the CFN scheduler. So all the code inside our handlers is being executed as many times as many stabilization steps it takes to complete the operation. An event chain would be marked as completed only if it succeeded. I.e. if we succeeded an attempt to create a resource with all the tags, it would be memorized in the `CallbackContext` and the next time this code is executed, it will simply be skipped (again, by the core library itself). If a chain fails, it won't be memorized as Uluru framework will decide it is a retriable error and the entire chain must be re-executed again. This is what we are trying to fix in this PR: if create with all tags fails, we attempt to create with system tags only. Assume this succeeds but the handler returns an in-progress event as we need to stabilize the new resource. Uluru scheduler will notice that and re-schedule another invocation of the handler lambda. The next invocation will execute this chain once again. As the create with all tags failed the previous time, it will attempt to invoke it once again. And once again it will fail. Then we will go to part2, where Uluru framework will realize that there is no need to re-create this resource and it is already being stabilized. We are trying to fix these constantly failing requests from part1 if part2 is stabilizing yet.

So, the essence of this PR is actually `isSoftFailTags` because it will prevent `part1` from being executed multiple times if it failed with `AccessDenied` once. If it failed for some other reasons and this is a retryable error, we should retry it once again.

`isAddTagsComplete` is a complement to this case that could easily be ignored by the downstream processor (i.e. there is nothing wrong in re-attaching existing tags to the resource: it is not an error, RDS will handle it nicely, but we can avoid it and save another call.

Hope this makes sense.

moataz-mhmd (approved):

Outige (approved):

PR #218, opened 2022-06-17
[OptionGroup] relax EngineName enum restrictions

Outige:
This commit introduces a series of changes that are aiming to restore the
backwards compatibility in the resource implementation.

Removed strict enum-based Engine type validation: using an enum enforces a case sensitive validation and breaks existing clients.

Signed-off-by: Tieg O'Sullivan [tiegaws@amazon.com](mailto:tiegaws@amazon.com)

### DIFF
--- a/aws-rds-optiongroup/aws-rds-optiongroup.json
+++ b/aws-rds-optiongroup/aws-rds-optiongroup.json
@@ -101,23 +101,7 @@
     },
     "EngineName": {
       "description": "Indicates the name of the engine that this option group can be applied to.",
-      "type": "string",
-      "enum": [
-        "aurora",
-        "aurora-mysql",
-        "aurora-postgresql",
-        "mariadb",
-        "mysql",
-        "oracle-ee",
-        "oracle-se2",
-        "oracle-se1",
-        "oracle-se",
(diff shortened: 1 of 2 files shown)


aws-rds-optiongroup/aws-rds-optiongroup.json
```diff
@@ -101,23 +101,7 @@
     },
     "EngineName": {
       "description": "Indicates the name of the engine that this option group can be applied to.",
```
khebul, line 103: Can we then put the link to [official RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateOptionGroup.html ) for reference?

osdrv (approved):

khebul (approved):

moataz-mhmd (approved):

aws-rds-optiongroup/aws-rds-optiongroup.json
osdrv: Will reflect this in the internal tracker. Generally speaking, we should revert this change when CFN starts supporting case insensitive enums (or if). Considering non-blocking.

PR #235, opened 2022-07-15
[OptionGroup] Add exception on requests that specify OptionGroupName

Outige:
This PR adds an exception to OptionGroup. When an OptionGroupName is specified.


aws-rds-optiongroup/src/test/java/software/amazon/rds/optiongroup/CreateHandlerTest.java
```diff
@@ -112,41 +119,31 @@ public void handleRequest_CreateSuccess() {
...
-        assertThat(response.getErrorCode()).isNull();
+        assertThat(response.getMessage()).isEqualTo("Encountered unsupported property OptionGroupName");
+        assertThat(response.getErrorCode()).isEqualTo(HandlerErrorCode.InvalidRequest);
 
-        verify(proxyClient.client(), times(1)).createOptionGroup(any(CreateOptionGroupRequest.class));
-        verify(proxyClient.client(), times(1)).describeOptionGroups(any(DescribeOptionGroupsRequest.class));
-        verify(proxyClient.client(), times(1)).listTagsForResource(any(ListTagsForResourceRequest.class));
+        verify(proxyClient.client(), times(0)).createOptionGroup(any(CreateOptionGroupRequest.class));
+        verify(proxyClient.client(), times(0)).describeOptionGroups(any(DescribeOptionGroupsRequest.class));
+        verify(proxyClient.client(), times(0)).listTagsForResource(any(ListTagsForResourceRequest.class));
```
osdrv, line 143: Nit: shall it be: `verifyNoMoreInteractions(rdsClient);`

osdrv (dismissed):

aws-rds-optiongroup/src/test/java/software/amazon/rds/optiongroup/CreateHandlerTest.java
```diff
@@ -72,10 +74,15 @@ public void setup() {
         proxy = new AmazonWebServicesClientProxy(logger, MOCK_CREDENTIALS, () -> Duration.ofSeconds(600).toMillis());
         rdsClient = mock(RdsClient.class);
         proxyClient = MOCK_PROXY(proxy, rdsClient);
+        setupResourceModelsForTests();
     }
 
     @AfterEach
-    public void tear_down() {
+    public void tear_down(TestInfo testInfo) {
+        if (testInfo.getTags().contains("IgnoreVerification") == true) {
```
khebul, line 82: nit: Can you make it into constant?

khebul (approved):

aws-rds-optiongroup/src/test/java/software/amazon/rds/optiongroup/AbstractTestBase.java
```diff
@@ -181,4 +173,8 @@ protected ProgressEvent<ResourceModel, CallbackContext> invokeHandleRequest(
     ) {
         return getHandler().handleRequest(getProxy(), request, context, getProxyClient(), logger);
     }
+
+    protected void setupResourceModelsForTests() {
```
khebul, line 177: Question: why do we need this function? My assumption is that by default option group name is null in the model. Is it incorrect?

Outige: My observation is that the default value is null. But it gets overridden by the random OptionGroupName generated by each individual test. Then subsequent calls using RESOURCE_MODEL fail due to there being an OptionGroupName present.

Please let me know if you have a better solution to get around this issue or if you have any follow up questions.

khebul: I have a slight concern that OptionGroup unit tests have side effects on Resource models shared by tests. This means that we are testing resource models that might differ from what we expect.

One option to fix this is to use a getter function that will create a new object every time it is invoked.

khebul (approved):

moataz-mhmd (approved):

PR #236, opened 2022-07-18
[DBParameterGroup] Make DBParameterGroup resource more resilient.

khebul:
This PR makes DBParameterGroup more resilient to transient failures, such as Throttling exceptions. This is achieved by saving intermediate results to Callback Context. The changes to callback context are saved on handler return if result of the execution was either `IN_PROGRESS` or `SUCCESS`. 

Previously the entire handler code was run in one execution of Lambda. This implementation breaks up execution of call chain into several stages. Call chain execution is broken up by setting minimal backoff period on `ProgressEvent`: some helper functions were added for that purpose.


aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/BaseHandlerStd.java
```diff
@@ -135,13 +135,17 @@ protected ProgressEvent<ResourceModel, CallbackContext> applyParametersWithReset
...
-        final Map<String, Parameter> defaultEngineParameters = Maps.newHashMap();
-        final Map<String, Parameter> currentDBParameters = Maps.newHashMap();
+        final Map<String, Parameter> defaultEngineParameters = callbackContext.getDefaultParameters();
+        final Map<String, Parameter> currentDBParameters = callbackContext.getCurrentParameters();
 
         return ProgressEvent.progress(model, callbackContext)
-                .then(progressEvent -> describeDefaultEngineParameters(progressEvent, defaultEngineParameters, proxy, proxyClient, null, requestLogger))
+                .then(progressEvent -> Commons.execAndSaveContextIfConditionIsNotMet(progressEvent,
+                        () -> describeDefaultEngineParameters(progressEvent, defaultEngineParameters, proxy, proxyClient, callbackContext.getDefaultParametersMarker(), requestLogger),
+                        CallbackContext::isDefaultParametersFetched))
```
osdrv, line 144: As describeDefaultEngineParameters alters the context already, we can blackbox the exec-once approach completely. This will eliminate the need for change here and below.

aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/UpdateHandler.java
```diff
@@ -41,6 +42,10 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
                 .build();
 
         return ProgressEvent.progress(request.getDesiredResourceState(), callbackContext)
+                .then(progress -> Commons.execOnceAndSaveContext(progress,
+                        () -> verifyDbParameterGroupExists(proxy, proxyClient, progress),
+                        CallbackContext::isVerifiedParameterGroupExists,
+                        CallbackContext::setVerifiedParameterGroupExists))
```
osdrv, line 48: Given this specific use-case pattern, I would suggest sticking to the framework primitives: a basic fetch call and a stabilization loop. It will provide the execOnce semantics from the box and do the fetch-and-proceed in a very unified way adopted across our resources. This will eliminate the need for the delayed progressEvent handling (`execOnceAndSaveContext`). I think this will simplify the change a bit.

aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/CreateHandler.java
```diff
@@ -61,7 +57,13 @@ private ProgressEvent<ResourceModel, CallbackContext> safeCreateDBParameterGroup
                                                                                      final ProgressEvent<ResourceModel, CallbackContext> progress,
                                                                                      final Tagging.TagSet allTags,
                                                                                      final RequestLogger requestLogger) {
-        final HandlerMethod<ResourceModel, CallbackContext> createMethod = (pxy, pcl, prg, tgs) -> createDBParameterGroup(pxy, pcl, prg, tgs, requestLogger);
+        final HandlerMethod<ResourceModel, CallbackContext> createMethod = (pxy, pcl, prg, tgs) ->
+                Commons.execOnceAndSaveContext(
+                        prg,
+                        () -> createDBParameterGroup(pxy, pcl, prg, tgs, requestLogger),
+                        CallbackContext::isParameterGroupCreated,
+                        CallbackContext::setParameterGroupCreated);
```
osdrv, line 65: Let's create and stabilize

PR #240, opened 2022-07-25
[Multiple Resources] Drop schema business logic validators

osdrv:
This commit drops some validators introduced earlier. The change includes enums, max values and specific engine types in DBCluster, DBInstance and EventSubscription.

The motivation for this change is to get rid of an overly restrictive validation on CFN end that might be out of sync with the RDS API region-specific configuration. The validators that touch business aspects of RDS turned out to be introducing more customer friction than help.

In a case an invalid attribute value is provided, CFN will act as a thin proxy and pass the value as-is to the RDS API. If the validation fails, RDS API will return an InvalidRequest exception which would be handled by the resource handler. RDS API messages seem to be more user-friendly as CFN auto-generates validation messages from the schema and they might seem cryptic.


aws-rds-dbcluster/aws-rds-dbcluster.json
```diff
@@ -178,8 +153,7 @@
     },
     "Port": {
       "description": "The port number on which the instances in the DB cluster accept connections. Default: 3306 if engine is set as aurora or 5432 if set to aurora-postgresql.",
-      "type": "integer",
-      "default": 3306
```
Outige, line 182: Why are we removing the default here?

What will the default be after this change?

aws-rds-dbcluster/aws-rds-dbcluster.json
```diff
@@ -256,51 +223,6 @@
...
-        "Engine"
-      ]
-    },
-    {
-      "anyOf": [
-        {
-          "anyOf": [
-            {
-              "required": [
-                "MasterUserPassword"
```
Outige, line 271: For attributes like this it might be nice to leave them in. The error message returned to the user is fairly straight forward. Indicating the user needs a MasterUserPassword for example.

Another nice benefit is this validation gets done very quickly and on the cfn end. If we remove this validation the customer has to wait for the call to RDS to validate.

aws-rds-dbcluster/aws-rds-dbcluster.json
osdrv: The engine-specific default value would be set by RDS API

aws-rds-dbcluster/aws-rds-dbcluster.json
osdrv: The motivation is to get rig of messages like `Properties validation failed for resource FooBarBaz with message: #: #: only 1 subschema matches out of 2 #/Iops: failed validation constraint for keyword [minimum]`

aws-rds-dbcluster/aws-rds-dbcluster.json
```diff
@@ -104,25 +102,11 @@
...
     "EngineMode": {
       "description": "The DB engine mode of the DB cluster, either provisioned, serverless, parallelquery, global, or multimaster.",
-      "type": "string",
-      "enum": [
-        "provisioned",
-        "serverless",
-        "parallelquery",
-        "global",
-        "multimaster"
-      ]
```
moataz-mhmd, line 125: I am only having two concerns
- This might effect public documentation as it is automatically generated from schema. So values could be removed from there. (We might reach out CFN team)
- Can effect CFN Stack designer tool users as it uses schema to provide autocomplete, validation,....

moataz-mhmd (approved):
I second this change as these enums can get out of sync easily and block customers stack for weeks waiting for fix deployment.

Outige (approved):

aws-rds-dbcluster/aws-rds-dbcluster.json
osdrv: Enums were never published before in the earlier versions of the schema. The documentation is quite static and human-written at the moment. We can add it later on if required.

PR #243, opened 2022-07-27
[DBInstance] Rework stabilization

osdrv:
This commit reimplements the current stabilization logic used in CreateHandler and UpdateHandler. The existing implementation is "shallow": it only checks the db instance status and ensures it reaches up "available" state.

The actual status is composite and is spread across multiple attributes. In particular, a complete availability status is defined as:
- db instance status is "available"
- all replication status info entries are in "replicating" status
- none of db parameter group entries is in "applying" status
- pending modified values is empty
- all vpc security groups are in "active" status
- all domain membership entries are either in "joined" or "kerberos-enabled" statuses
- promotion tier is updated

This commit alters the stabilization logic so it conforms to the contract above. This approach also invalidates the need for n-probing: a workaround introduced earlier in order to extrapolate the composite availability status with a simplistic n-times check.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -11,7 +11,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.Function;
-import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 import org.apache.commons.lang3.BooleanUtils;
@@ -50,6 +49,7 @@ import software.amazon.awssdk.services.rds.model.InvalidRestoreException;
 import software.amazon.awssdk.services.rds.model.InvalidVpcNetworkStateException;
 import software.amazon.awssdk.services.rds.model.KmsKeyNotAccessibleException;
 import software.amazon.awssdk.services.rds.model.OptionGroupMembership;
+import software.amazon.awssdk.services.rds.model.PendingModifiedValues;
 import software.amazon.awssdk.services.rds.model.ProvisionedIopsNotAvailableInAzException;
 import software.amazon.awssdk.services.rds.model.SnapshotQuotaExceededException;
 import software.amazon.awssdk.services.rds.model.StorageQuotaExceededException;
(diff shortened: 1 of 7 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -96,10 +102,13 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
     );
 
     protected final static HandlerConfig DEFAULT_DB_INSTANCE_HANDLER_CONFIG = HandlerConfig.builder()
```
khebul, line 104: Is it possible to dynamically set config timeout based on IAM role provided?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -497,7 +474,58 @@ protected boolean isDbInstanceStabilized(
...
+        return (pending == null) || (pending.dbInstanceClass() == null &&
+                pending.allocatedStorage() == null &&
+                pending.caCertificateIdentifier() == null &&
+                pending.masterUserPassword() == null &&
+                pending.port() == null &&
+                pending.backupRetentionPeriod() == null &&
+                pending.multiAZ() == null &&
+                pending.engineVersion() == null &&
+                pending.iops() == null &&
+                pending.dbInstanceIdentifier() == null
```
moataz-mhmd, line 514: What about remaining features? Should we add them?
```
pending.dbInstanceClass()
pending.licenseModel()
pending.storageType()
pending.caCertificateIdentifier()
pending.dbSubnetGroupName()
pending.pendingCloudwatchLogsExports()
pending.hasProcessorFeatures() ? this.processorFeatures() : null
pending.iamDatabaseAuthenticationEnabled()
pending.automationModeAsString()
pending.resumeFullAutomationModeTime()
```

moataz-mhmd (dismissed):

Outige (dismissed):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -85,8 +85,14 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
 
     public static final String API_VERSION_V12 = "2012-09-17";
 
-    public static final String PENDING_REBOOT_STATUS = "pending-reboot";
-    public static final String IN_SYNC_STATUS = "in-sync";
+    static final String DB_PARAMETER_GROUP_STATUS_APPLYING = "applying";
+    static final String IN_SYNC_STATUS = "in-sync";
+    static final String PENDING_REBOOT_STATUS = "pending-reboot";
```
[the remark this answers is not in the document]
osdrv: Semantically, no: they represent particular statuses concerning distinct resources. Hence, we would need to introduce the same amount of independent enums representing transitions in their own domain. As these transitions are out of the scope of our interests in the context of stabilization, we simply don't care about other statuses. Hence, a set of plain strings.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
osdrv: This set was takes from the legacy implementation. It won't hurt to add them, I guess. On it.

Outige (approved):

moataz-mhmd (approved):

khebul (approved):

PR #245, opened 2022-07-27
[DBInstance] Do not set DBParameterGroup on create of read replica

khebul:
Not all engines support setting DBParameterGroup on creation of read replicas. Instead, we set read replicas on subsequent modify request.


moataz-mhmd (approved):

moataz-mhmd (approved):
If It is handy to you, Can you refer what is the exception thrown in case engine does not support setting DPParameterGroup while creating read replica.

osdrv (approved):

PR #248, opened 2022-07-29
[DBInstance][DBCluster] set timeout for Create and Update handlers to 36 hours

khebul:
In the recent changes the timeout for Create and Update stabilization for `DBInstance` and `DBCluster` resources was set to 36 hours, however the handler timeout was not updated to match. This PR addresses this oversight.

### DIFF
--- a/aws-rds-dbcluster/resource-role.yaml
+++ b/aws-rds-dbcluster/resource-role.yaml
@@ -7,7 +7,7 @@ Resources:
   ExecutionRole:
     Type: AWS::IAM::Role
     Properties:
-      MaxSessionDuration: 8400
+      MaxSessionDuration: 43200
       AssumeRolePolicyDocument:
         Version: '2012-10-17'
         Statement:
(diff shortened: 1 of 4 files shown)


aws-rds-dbcluster/resource-role.yaml
```diff
@@ -7,7 +7,7 @@ Resources:
   ExecutionRole:
     Type: AWS::IAM::Role
     Properties:
-      MaxSessionDuration: 8400
+      MaxSessionDuration: 43200
```
Outige, line 10: This is 720 minutes?
Why is this change being made?

Outige (approved):

aws-rds-dbcluster/resource-role.yaml
khebul: This is max session duration for IAM role in seconds. This role is being assumed when handlers is executed. The value is set to the [maximum allowed value supported by IAM role](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html).

This file is automatically generated by CloudFomation tooling based on a [resource schema](https://github.com/aws-cloudformation/cloudformation-resource-schema/blob/eeb4064ecff6ebfab325b02096b0bcc83d5d848a/src/main/resources/schema/provider.definition.schema.v1.json#L18-L23). The value for timeout is within the limits provided by CFN platform and the expectation is for CFN platform to manage the sessions to support this value.

osdrv (approved):

PR #251, opened 2022-08-02
[DBInstance] Change stabilization logic on update

osdrv:
This commit introduces a distinction between stabilization processes after create and update. The change is introduced for the sake of a complete logic parity with the historical implementation. In particular, the post-update stabilization is:
- DBInstance status is `available`
- DBParameterGroup status is `in-sync`
- OptionGroup status is `in-sync`
- if DBInstance is a member of a DBCluster:
  - DBClusterParameterGroup is `in-sync`

Both stabilizers perform an additional assertion on DBInstance status. The assertion code performs an early process interruption if DBInstance enters one of the following statuses:
- failed
- incompatible-restore
- incompatible-network
- incompatible-parameters
- inaccessible-encryption-credentials

In this case the handler will throw a `CfnNotStabilizedException`.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -21,9 +21,7 @@ import software.amazon.awssdk.services.ec2.model.DescribeSecurityGroupsResponse;
 import software.amazon.awssdk.services.ec2.model.SecurityGroup;
 import software.amazon.awssdk.services.rds.RdsClient;
 import software.amazon.awssdk.services.rds.model.DBCluster;
-import software.amazon.awssdk.services.rds.model.DBClusterMember;
 import software.amazon.awssdk.services.rds.model.DBInstance;
-import software.amazon.awssdk.services.rds.model.DBParameterGroupStatus;
 import software.amazon.awssdk.services.rds.model.DBSnapshot;
 import software.amazon.awssdk.services.rds.model.DbInstanceAlreadyExistsException;
 import software.amazon.awssdk.services.rds.model.DbInstanceAutomatedBackupQuotaExceededException;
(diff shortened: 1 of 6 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -460,31 +458,63 @@ protected boolean isDbInstanceDeleted(
...
+        assertNoTerminalStatus(dbInstance);
+
         return false;
     }
 
-    protected boolean isDbInstanceStabilized(
+    private void assertNoTerminalStatus(final DBInstance dbInstance) throws CfnNotStabilizedException {
+        final DBInstanceStatus status = DBInstanceStatus.fromString(dbInstance.dbInstanceStatus());
+        if (status != null && status.isTerminal()) {
+            throw new CfnNotStabilizedException(new Exception("DB Instance is in state: " + status.toString()));
```
khebul, line 477: nit: This does not match the rest of the code that we use for returning values to CFN framework. As far as I can tell, throwing `CfnNotStabilizedException` is equivalent of returning failed result with an error code `HandlerErrorCode.NotStabilized`. 

Do you think using the same error handling style is going to introduce more complexity?

osdrv: I totally agree with you: we've been sticking to returning a faulty event back to the handler. The reason this approach is used here is a lack of an interruption mechanism in the stabilizer (it seems):
```java
/**
  * @param callback, the stabilize predicate that is called several times to
  *            determine success.
  * @return true if the condition of stabilization has been meet else false.
  */
```
Thinking of an alternative, we can throw some other exception here and catch it outside the code block that performs stabilization. Having that said, i believe we will have quite some more code blocks wrapped in try-catch blocks that would make it look clumsy. To me this shortcut looks like a fair balance between this bulky try-catch party and an oopsie-doopsie-we-got-another-exit-point situation. Please let me know if it makes sense.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -460,31 +458,63 @@ protected boolean isDbInstanceDeleted(
...
+        DBInstance dbInstance;
         try {
-            fetchDBInstance(rdsProxyClient, model);
+            dbInstance = fetchDBInstance(rdsProxyClient, model);
         } catch (DbInstanceNotFoundException e) {
             // the instance is gone, exactly what we need
             return true;
         }
+
+        assertNoTerminalStatus(dbInstance);
```
Outige, line 469: I really like this line

Outige (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
khebul: Yeah, looks like an oversight in stabilization mechanism. Let's keep CFN exception for now 👍

khebul (approved):

PR #253, opened 2022-08-02
[DBInstance] Unify stabilization for create and update

osdrv:
This PR is a follow-up on the change introduced in https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/251.

The initial change introduced a distinction between create and update stabilization logic. This distinction was not precise enough: a DBInstance can enter pending-reboot state after an update and the existing logic would block as it requires an in-sync state.

This commit introduces a change to the stabilization logic. Any mutating change would be stabilized with the same logic (including creates and updates) and post-reboot gets its own stabilizer that would ensure the minimal set of availability flags and assert an in-sync (rather than "anything but applying") status.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -335,7 +335,7 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
                         modifyRequest,
                         proxyInvocation.client()::modifyDBInstance
                 ))
-                .stabilize((modifyRequest, response, proxyInvocation, model, context) -> isDBInstanceStabilizedAfterUpdate(proxyInvocation, model))
+                .stabilize((modifyRequest, response, proxyInvocation, model, context) -> isDBInstanceStabilizedAfterMutate(proxyInvocation, model))
                 .handleError((modifyRequest, exception, client, model, context) -> Commons.handleException(
                         ProgressEvent.progress(model, context),
                         exception,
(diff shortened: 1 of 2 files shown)


Outige (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -384,7 +384,7 @@ protected ProgressEvent<ResourceModel, CallbackContext> awaitDBInstanceAvailable
                 .translateToServiceRequest(Function.identity())
                 .backoffDelay(config.getBackoff())
                 .makeServiceCall(NOOP_CALL)
-                .stabilize((request, response, proxyInvocation, model, context) -> isDBInstanceStabilizedAfterUpdate(proxyInvocation, model))
+                .stabilize((request, response, proxyInvocation, model, context) -> isDBInstanceStabilizedAfterReboot(proxyInvocation, model))
```
khebul, line 387: nit: method invoking this call is called `awaitDBInstanceAvailableStatus`, which might confuse us in the future. Maybe it is worth indicating that this method should only be called after reboot?

khebul (approved):

moataz-mhmd (dismissed):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
osdrv: Good call. Fixed now.

moataz-mhmd (approved):

Outige (approved):

PR #256, opened 2022-08-04
[Common] Implement custom UserAgent signature

osdrv:
This commit introduces a custom user agent signature to RDS CFN resource handler API invocations. The motivation for the change is to distinguish API calls made by an RDS CFN handler and any other generic API calls performed by the Java SDK.

The commit introduces a new `BaseSdkClientProvider`. Resource client providers are supposed to extend this abstract class and implement their custom `httpClient` decoration logic.

We introduce `RdsUserAgentProvider` class that is supposed to provide the corresponding user agent signature. The current implementation provides a static user agent signature prefix: `AWS/CloudFormation/RDS`. Maven build process was extended with git-commit-id-maven-plugin which creates version.properties file in the classpath upon a build. The user agent signature is being annotated with the corresponding git commit abbreviation.

---


aws-rds-cfn-common/src/main/java/software/amazon/rds/common/client/RdsUserAgentProvider.java
```diff
@@ -0,0 +1,47 @@
+package software.amazon.rds.common.client;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+import com.amazonaws.util.StringUtils;
+import software.amazon.rds.common.test.ExcludeFromJacocoGeneratedReport;
+
+public final class RdsUserAgentProvider {
```
moataz-mhmd, line 10: Is it possible to add date of PR merge or any relatable date so that operator can compare versions based on these dates of changes?

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/client/RdsUserAgentProvider.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static final String SDK_CLIENT_USER_AGENT_PREFIX = "AWS/CloudFormation/RDS";
+
+    static final String VERSION_PROPERTIES_FILE_PATH = "version.properties";
+    static final String GIT_COMMIT_ID_ABBREV_PROPERTY = "git.commit.id.abbrev";
```
moataz-mhmd, line 15: Is it possible to add first 50/60 characters of git commit message? I think it might boost readability of UserAgent signature to operator that he do not need to return to git history?

moataz-mhmd (approved):
Thanks for implementing it as it will make investigations easier.

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/client/RdsUserAgentProvider.java
osdrv: No, we can only use information from the SCM (git). A PR, its merge data etc is coming from GitHub, which is a side 3rd party dependency in this context, hence should be avoided as a part of the internal build process.

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/client/RdsUserAgentProvider.java
osdrv: Given a commit message is a free-format UTF8 text, I won't really trust it: we can introduce more harm than help. any SCM is able to identify a commit by the first 4-6 characters of its SHA hash, given it has a high hash entropy. E.g. here Github finds my commit by the first 5 characters: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/commit/11210.

khebul (dismissed):

Outige (dismissed):

moataz-mhmd (dismissed):

khebul (dismissed):

Outige (dismissed):

moataz-mhmd (approved):

Outige (approved):

khebul (approved):

PR #262, opened 2022-08-12
[DBInstance] No `vpcSecurityGroupIds` for dbcluster instance modifications

osdrv:
This commit eliminates `vpcSecurityGroupIds` setting for `ModifyDBInstance` requests if an instance is a member of a DBCluster.

The reason for this change is to avoid a rejection by RDS API as DBCluster instance vpc security groups are managed by the cluster object, not atomic instances.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
@@ -303,14 +303,18 @@ public class Translator {
                 .promotionTier(desiredModel.getPromotionTier())
                 .storageType(desiredModel.getStorageType())
                 .tdeCredentialArn(desiredModel.getTdeCredentialArn())
-                .tdeCredentialPassword(desiredModel.getTdeCredentialPassword())
-                .vpcSecurityGroupIds(desiredModel.getVPCSecurityGroups());
+                .tdeCredentialPassword(desiredModel.getTdeCredentialPassword());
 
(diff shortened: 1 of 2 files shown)


khebul (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -303,15 +303,19 @@ public static ModifyDbInstanceRequest modifyDbInstanceRequest(
...
                 Optional.ofNullable(previousModel).map(ResourceModel::getEnableCloudwatchLogsExports).orElse(Collections.emptyList()),
                 desiredModel.getEnableCloudwatchLogsExports()
         );
         builder.cloudwatchLogsExportConfiguration(cloudwatchLogsExportConfiguration);
 
+        // VPC security groups can not be assigned directly on an Aurora instance.
+        if (StringUtils.isBlank(desiredModel.getDBClusterIdentifier())) {
+            builder.vpcSecurityGroupIds(desiredModel.getVPCSecurityGroups());
+        }
+
```
moataz-mhmd, line 318: I am afraid this can be misleading in case of customer specified `DBClusterIdentifier` and `vpcSecurityGroupIds` in update. `vpcSecurityGroupIds` will be ignored and update succeed instead of failure as customer can not update `vpcSecurityGroupIds` if db instance is part of cluster.

moataz-mhmd (approved):

PR #264, opened 2022-08-12
[DBInstance] Unset empty `vpcSecurityGroupIds`

osdrv:
This commit changes the way Translator code sets `vpcSecurityGroupIds` attribute on the following requests:
  - `CreateDBInstance`
  - `CreateDBInstanceReadReplica`
  - `RestoreDBIntanceFromDBSnapshot`
  - `ModifyDBInstance`

The motivation for this change is to ensure no explicit empty list is being set as in this case RDS API won't recognize it as a no-change but will interpret it as an explicitly empty collection and fail if the configuration is incorrect.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
@@ -13,6 +13,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.BooleanUtils;
 
 import software.amazon.awssdk.services.ec2.model.DescribeSecurityGroupsRequest;
@@ -94,7 +95,7 @@ public class Translator {
                 .storageType(model.getStorageType())
                 .tags(Tagging.translateTagsToSdk(tagSet))
                 .useDefaultProcessorFeatures(model.getUseDefaultProcessorFeatures())
-                .vpcSecurityGroupIds(model.getVPCSecurityGroups())
+                .vpcSecurityGroupIds(CollectionUtils.isNotEmpty(model.getVPCSecurityGroups()) ? model.getVPCSecurityGroups() : null)
                 .build();
     }
 
@@ -146,7 +147,7 @@ public class Translator {
                 .tdeCredentialArn(model.getTdeCredentialArn())
                 .tdeCredentialPassword(model.getTdeCredentialPassword())
                 .useDefaultProcessorFeatures(model.getUseDefaultProcessorFeatures())
-                .vpcSecurityGroupIds(model.getVPCSecurityGroups())
+                .vpcSecurityGroupIds(CollectionUtils.isNotEmpty(model.getVPCSecurityGroups()) ? model.getVPCSecurityGroups() : null)
                 .build();
     }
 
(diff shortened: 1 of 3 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -146,7 +147,7 @@ public static RestoreDbInstanceFromDbSnapshotRequest restoreDbInstanceFromSnapsh
                 .tdeCredentialArn(model.getTdeCredentialArn())
                 .tdeCredentialPassword(model.getTdeCredentialPassword())
                 .useDefaultProcessorFeatures(model.getUseDefaultProcessorFeatures())
-                .vpcSecurityGroupIds(model.getVPCSecurityGroups())
+                .vpcSecurityGroupIds(CollectionUtils.isNotEmpty(model.getVPCSecurityGroups()) ? model.getVPCSecurityGroups() : null)
```
moataz-mhmd, line 150: Unfornately, putting `vpcSecurityGroupIds` with null will be converted to empty list as well.

here is the implementation
```
        public final ModifyDbInstanceRequest.Builder vpcSecurityGroupIds(Collection<String> vpcSecurityGroupIds) {
            this.vpcSecurityGroupIds = VpcSecurityGroupIdListCopier.copy(vpcSecurityGroupIds);
            return this;
        }
```
and
```
final class VpcSecurityGroupIdListCopier {
    VpcSecurityGroupIdListCopier() {
    }

    static List<String> copy(Collection<String> vpcSecurityGroupIdListParam) {
        Object list;
        if (vpcSecurityGroupIdListParam != null && !(vpcSecurityGroupIdListParam instanceof SdkAutoConstructList)) {
            List<String> modifiableList = new ArrayList();
            vpcSecurityGroupIdListParam.forEach((entry) -> {
                modifiableList.add(entry);
            });
            list = Collections.unmodifiableList(modifiableList);
        } else {
            list = DefaultSdkAutoConstructList.getInstance();
        }

        return (List)list;
    }
}
```

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -304,7 +305,7 @@ public static ModifyDbInstanceRequest modifyDbInstanceRequest(
                 .storageType(desiredModel.getStorageType())
                 .tdeCredentialArn(desiredModel.getTdeCredentialArn())
                 .tdeCredentialPassword(desiredModel.getTdeCredentialPassword())
-                .vpcSecurityGroupIds(desiredModel.getVPCSecurityGroups());
+                .vpcSecurityGroupIds(CollectionUtils.isNotEmpty(desiredModel.getVPCSecurityGroups()) ? desiredModel.getVPCSecurityGroups() : null);
```
moataz-mhmd, line 308: Should we add same behavior to `modifyDbInstanceRequestV12`?

osdrv: No: it does not support vpc security groups

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
osdrv: I know, this is the idea. The `SdkAutoConstructList` is getting serialized into null rather than an empty list (like `ArrayList`)

moataz-mhmd (approved):

khebul (approved):

PR #271, opened 2022-08-17
[DBInstance] Enforce final snapshot if not explicitly requested

osdrv:
This commit enforces the final snapshot creation on DBInstance delete in case the request contains no data on whether a final snapshot creation is requested or not. The reason for this change is the referential value type returned by `ResourceHandlerRequest.getSnapshotRequested()`. The return value is a Boolean, meaning it can be yes/no/null(no data). This commit addresses the latter case where the invocation request is neither asking for a final snapshot nor disclaiming it.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
@@ -49,7 +49,7 @@ public class DeleteHandler extends BaseHandlerStd {
         // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html
         // For AWS::RDS::DBInstance resources that don't specify the DBClusterIdentifier property, the default policy is Snapshot.
         // Final snapshots are not allowed for read replicas and cluster instances.
-        if (BooleanUtils.isTrue(request.getSnapshotRequested()) &&
+        if (BooleanUtils.isNotFalse(request.getSnapshotRequested()) &&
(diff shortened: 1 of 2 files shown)


khebul (approved):

Outige (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/DeleteHandler.java
```diff
@@ -49,7 +49,7 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
         // https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html
         // For AWS::RDS::DBInstance resources that don't specify the DBClusterIdentifier property, the default policy is Snapshot.
         // Final snapshots are not allowed for read replicas and cluster instances.
-        if (BooleanUtils.isTrue(request.getSnapshotRequested()) &&
+        if (BooleanUtils.isNotFalse(request.getSnapshotRequested()) &&
```
moataz-mhmd, line 52: Just to put on radar. This will make contract tests leave snapshot behind whenever they run into any AWS account. THese snapshot will sum up till quota is full ~50 snapshots then contract test will start to fail. Unfortunately there is no way around to write the parameter in contract test request.

osdrv: Ack, many thanks for raising it. We will solve it progressively. I believe we can justify a temporary cleanup TOIL by loosing a risk of an accidental snapshot removal.

PR #277, opened 2022-08-19
[DBInstance] Add ResourceHandlerRequest validation

osdrv:
This commit adds a request validation layer to the handler execution chain. The motivation is to extend the schema-based validation with a range of more flexible and more dynamic validation methods.

The change introduces a new class `ValidatedRequest` that extends `ResourceHandlerRequest` with 0 overloading: it is here to distinguish a validated request from a raw one.

The change adds a new validation on the `sourceRegion` attribute. If a provided region is incorrect, the validation will refuse such a request well before it reaches the SDK endpoint connection initialization.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -6,16 +6,19 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apache.commons.lang3.BooleanUtils;
 
 import com.amazonaws.util.CollectionUtils;
+import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.ec2.Ec2Client;
 import software.amazon.awssdk.services.ec2.model.DescribeSecurityGroupsResponse;
 import software.amazon.awssdk.services.ec2.model.SecurityGroup;
@@ -78,6 +81,8 @@ import software.amazon.rds.dbinstance.client.ApiVersionDispatcher;
 import software.amazon.rds.dbinstance.client.Ec2ClientProvider;
 import software.amazon.rds.dbinstance.client.RdsClientProvider;
 import software.amazon.rds.dbinstance.client.VersionedProxyClient;
+import software.amazon.rds.dbinstance.request.RequestValidationException;
+import software.amazon.rds.dbinstance.request.ValidatedRequest;
 import software.amazon.rds.dbinstance.util.ResourceModelHelper;
 
 public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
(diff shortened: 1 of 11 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -295,20 +304,54 @@ protected ApiVersionDispatcher<ResourceModel, CallbackContext> getApiVersionDisp
...
-            final Logger logger);
+            final Logger logger
+    ) {
+        try {
+            validateRequest(request);
+        } catch (RequestValidationException exception) {
+            return ProgressEvent.defaultFailureHandler(exception, HandlerErrorCode.InvalidRequest);
+        }
+
+        return handleRequest(proxy, new ValidatedRequest<ResourceModel>(request), context, rdsProxyClient, ec2ProxyClient, logger);
```
moataz-mhmd, line 345: Could you elaborate more what is the benefit of `ValidatedRequest` class? I see no extra logic there and validation logic already happened in 
```
             validateRequest(request);
```

moataz-mhmd (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
osdrv: It is a pure semantic container class. No specific purpose rather than making it explicit that at this stage the validation has been completed. Overloading the corresponding `handleRequest` makes it explicit. In reality, I had some other less conceptual and more down-to-earth reasons. I needed to ensure that the entrypoint to `handleRequest` is sort of guarded. No matter what method overloading I was sketching out, it was clear that the invoker (most likely a test code) still has a way to unconsciously avoid the from-now-on-mandatory request validation. I.e. I needed a more reliable mechanism in place that would ensure the request validation but won't rely on "please call me instead of anything else" approach. Introducing an overloaded class sounded like a decent solution: by a price of an additional object construction we are isolating 2 stages of the request execution: the "unvalidated" and "validated". This also allowed to me make the change with a minimum amount of modifications to the test code. Hope this makes sense.

Outige (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/request/ValidatedRequest.java
```diff
@@ -0,0 +1,28 @@
+package software.amazon.rds.dbinstance.request;
+
+import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
+
+public class ValidatedRequest<T> extends ResourceHandlerRequest<T> {
```
khebul, line 5: I have a concern regarding this class. If `ResourceHandlerRequest` gets updated in the wrapper package, we will have to modify this class. There is a high chance that such changes will be missed.

osdrv: `ResourceHandlerRequest` is a `@Data` class with an auto-generated builder, all-args constructor and no-args constructor. `ValidatedRequest` exposes only 1 public constructor method, which proxies to the all-args constructor. Any change to `ResourceHandlerRequest` will immediately break the all-args constructor (assuming there is any (literally any) change in the order of the arguments or their set. This is the reason I used all-args constructor: to ensure any external change would break the compilation.

PR #285, opened 2022-08-24
[DBParameterGroup] Filtered `DescribeDBParameters`

osdrv:
This commit adds an additional filtering to `DescribeDBParameters`. Having it in-place reduces the page depth of the cluster parameter fetch. The amount of parameters to be fetched is limited to the parameters that are directly enumerated in the ResourceModel.

Relevant PR: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/284


moataz-mhmd (changes requested):
I am afraid this optimization will not work in DBParameterGroup resource and can be customer impacting. Resource handler need to reset parameters that not having default values when customer removes it from stack.
For example,
1. Customer provided stack 
```
...
      Family: mysql5.7
      Parameters:
        character_set_server: utf8mb4 #<<<<<<<<<<<<
        collation_server: "utf8mb4_unicode_ci" #<<<<<<<<<<<
        innodb_buffer_pool_instances: 16
        binlog_format: "ROW"
        event_scheduler: 'ON'
      Description:
...
```
2. Customer submitted update stack 
```
...
      Family: mysql5.7
      Parameters:
        character_set_server: utf8mb4 #<<<<<<<<<<<<
        collation_server: "utf8mb4_unicode_ci" #<<<<<<<<<<<
        innodb_buffer_pool_instances: 16
        binlog_format: "ROW"
      Description:
...
```
Note here that customer removed `event_scheduler` from stack and that's mean resource handler need to reset this parameter to its default value `OFF` again with `ResetParameters` API call.
When customer removed `event_scheduler` from stack, It is no longer in `model.parameter()` so they will not be checked in `DefaultDescribe` and regular `Describe`
then will be totally ignored which can be problematic as we need to describe then reset it if not having default value specifically.

Unfortunately, Same applies for **out of band customer changes** so we can not relay on `previousModel` parameters as well to avoid describe all parameters.

> **_NOTE:_** 
Out of band change means customer can change his parameter group directly from rds client without CFN knowing about it.

you can find DBParameterGroup resetting code here: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/cdfc824b2a4821df30be9c1513290d5451eb3a7d/aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/BaseHandlerStd.java#L174-L189


#### Why this change works in DBClusterParameterGroup?
Because in this resource we reset all parameters generally without specifing in parameters names here: 

https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java#L258-L270

#### Why `DBParameterGroup` resource do not have same behavior and resetting all parameters instead to reset specific parameters by name?
Because resetting all parameters can trigger unnecessary db instance reboot.

**TDLR: I would recommend go only with https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/284 optimization and discard/close this PR for DBParameterGroup**

osdrv:
@moataz-mhmd thanks very much for your extended comment. Any chance you could share an example of the stack that will reproduce this case?

moataz-mhmd:
Yes, Here is a scenario that can triggers this issue you can check

```
AWSTemplateFormatVersion: 2010-09-09
Description: RDS DBParameterGroup Canary

Resources:
  RDSDBParameterGroup:
    Type: AWS::RDS::DBParameterGroup
    Properties:
      Parameters:
        authentication_timeout: 300
        autovacuum_analyze_scale_factor: 40
      Family: postgres13
      Description: test
```

Then do update on stack by removing `autovacuum_analyze_scale_factor` parameter
```

AWSTemplateFormatVersion: 2010-09-09
Description: RDS DBParameterGroup Canary

Resources:
  RDSDBParameterGroup:
    Type: AWS::RDS::DBParameterGroup
    Properties:
      Parameters:
        authentication_timeout: 300
      Family: postgres13
      Description: test
```
Then check ParameterGroup in you account after update. You will find `autovacuum_analyze_scale_factor` still 40 but it should be reset to its default value 0.05.

khebul (approved):

Outige (approved):

moataz-mhmd (commented):
Thanks for addressing.

moataz-mhmd (approved):

PR #292, opened 2022-08-29
[DBClusterParameterGroup] changing stabilization from describe cluste…

moataz-mhmd:
*Description of changes:*
- Remove stabilization on DBClusters and replace with 5 min for waiting propagation according to https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyDBClusterParameterGroup.html.

### DIFF
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
@@ -76,10 +76,9 @@ public final class Commons {
             final VoidBiFunction<C, Boolean> conditionSetter
     ) {
         if (!conditionGetter.apply(progress.getCallbackContext())) {
-            return func.enact().then(p -> {
-                conditionSetter.apply(p.getCallbackContext(), true);
-                return p;
-            });
+            ProgressEvent<M, C> progressEvent = func.enact();
+            conditionSetter.apply(progressEvent.getCallbackContext(), true);
+            return progressEvent;
         }
         return progress;
     }
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
@@ -1,5 +1,6 @@
 package software.amazon.rds.common.handler;
 
+import java.time.Duration;
 import java.util.function.Function;
 
 import software.amazon.awssdk.core.exception.SdkClientException;
@@ -83,5 +84,12 @@ public final class Commons {
             });
         }
         return progress;
+    }
+
+    public static <M, C> ProgressEvent<M, C> sleep(
+            final ProgressEvent<M, C> progress,
+            final Duration delay
+    ) {
+        return ProgressEvent.defaultInProgressHandler(progress.getCallbackContext(), Math.toIntExact(delay.getSeconds()), progress.getResourceModel());
     }
 }
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
@@ -21,7 +21,8 @@ public final class Commons {
             .withErrorCodes(ErrorStatus.failWith(HandlerErrorCode.AccessDenied),
                     ErrorCode.AccessDenied,
                     ErrorCode.AccessDeniedException,
-                    ErrorCode.NotAuthorized)
+                    ErrorCode.NotAuthorized,
+                    ErrorCode.UnauthorizedOperation)
             .withErrorCodes(ErrorStatus.failWith(HandlerErrorCode.Throttling),
                     ErrorCode.ThrottlingException,
                     ErrorCode.Throttling)
@@ -76,9 +77,10 @@ public final class Commons {
             final VoidBiFunction<C, Boolean> conditionSetter
     ) {
         if (!conditionGetter.apply(progress.getCallbackContext())) {
-            ProgressEvent<M, C> progressEvent = func.enact();
-            conditionSetter.apply(progressEvent.getCallbackContext(), true);
-            return progressEvent;
+            return func.enact().then(p -> {
+                conditionSetter.apply(p.getCallbackContext(), true);
+                return p;
+            });
         }
         return progress;
     }
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
@@ -2,10 +2,8 @@ package software.amazon.rds.common.handler;
 
 import java.time.Duration;
 
-import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
-import lombok.NoArgsConstructor;
 import lombok.experimental.SuperBuilder;
 import software.amazon.cloudformation.proxy.delay.Constant;
 
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
@@ -2,11 +2,14 @@ package software.amazon.rds.common.handler;
 
 import java.time.Duration;
 
+import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.experimental.SuperBuilder;
 import software.amazon.cloudformation.proxy.delay.Constant;
 
-@Builder
+@SuperBuilder
 public class HandlerConfig {
 
     @Getter
@@ -19,8 +22,4 @@ public class HandlerConfig {
             .delay(Duration.ofSeconds(30))
             .timeout(Duration.ofMinutes(90))
             .build();
-
-    @Getter
-    @Builder.Default
-    final private int stabilizationInSeconds = 60;
 }
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
@@ -19,4 +19,8 @@ public class HandlerConfig {
             .delay(Duration.ofSeconds(30))
             .timeout(Duration.ofMinutes(90))
             .build();
+
+    @Getter
+    @Builder.Default
+    final private int stabilizationInSeconds = 60;
 }
--- a/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
+++ b/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
@@ -37,7 +37,6 @@ import software.amazon.rds.common.error.ErrorCode;
 import software.amazon.rds.common.error.ErrorRuleSet;
 import software.amazon.rds.common.error.ErrorStatus;
 import software.amazon.rds.common.handler.Commons;
-import software.amazon.rds.common.handler.HandlerConfig;
 import software.amazon.rds.common.handler.Tagging;
 import software.amazon.rds.common.logging.LoggingProxyClient;
 import software.amazon.rds.common.logging.RequestLogger;
@@ -50,7 +49,7 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
     protected static final String STACK_NAME = "rds";
     protected static final int MAX_LENGTH_GROUP_NAME = 255;
     // 5 min for waiting propagation according to https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyDBClusterParameterGroup.html
-    protected static final int STABILIZATION_DELAY_SECONDS = 5 * 60;
+    protected static final Duration STABILIZATION_DELAY = Duration.ofMinutes(5);
     protected static final int MAX_PARAMETERS_PER_REQUEST = 20;
 
     protected static final ErrorRuleSet DEFAULT_DB_CLUSTER_PARAMETER_GROUP_ERROR_RULE_SET = ErrorRuleSet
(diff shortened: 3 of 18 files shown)


aws-rds-dbclusterparametergroup/src/test/java/software/amazon/rds/dbclusterparametergroup/CreateHandlerTest.java
```diff
@@ -365,7 +306,7 @@ public void handleRequest_SuccessWithParameters() {
 
         verify(proxyRdsClient.client()).createDBClusterParameterGroup(any(CreateDbClusterParameterGroupRequest.class));
         verify(proxyRdsClient.client()).modifyDBClusterParameterGroup(any(ModifyDbClusterParameterGroupRequest.class));
-        verify(proxyRdsClient.client()).describeDBClustersPaginator(any(DescribeDbClustersRequest.class));
+//        verify(proxyRdsClient.client()).describeDBClustersPaginator(any(DescribeDbClustersRequest.class));
```
khebul, line 309: nit: commented out code

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
```diff
@@ -62,7 +62,7 @@ public static <M, C> ProgressEvent<M, C> handleException(
             // If we not set model and context to null, CFN will attempt to perform delete of the resource it has attempted to create,
             // even if resource belongs to a different stack altogether, or was created out of bounds
             if (handlerErrorStatus.getHandlerErrorCode() == HandlerErrorCode.AlreadyExists) {
-                return ProgressEvent.failed(null, null, handlerErrorStatus.getHandlerErrorCode(), exception.getMessage());
+                return ProgressEvent.failed(null, progress.getCallbackContext(), handlerErrorStatus.getHandlerErrorCode(), exception.getMessage());
```
khebul, line 65: Please update the comment above to reflect the change.

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
```diff
@@ -77,10 +77,9 @@ public static <M, C> ProgressEvent<M, C> execOnce(
             final VoidBiFunction<C, Boolean> conditionSetter
     ) {
         if (!conditionGetter.apply(progress.getCallbackContext())) {
-            return func.enact().then(p -> {
-                conditionSetter.apply(p.getCallbackContext(), true);
-                return p;
-            });
+            ProgressEvent<M, C> progressEvent = func.enact();
+            conditionSetter.apply(progressEvent.getCallbackContext(), true);
```
khebul, line 81: In case then the `func` is executed more than once (e.g. stabilisation) this will cause this code to exit prematurely. I think it is safer to throw if progress delay is not zero.

moataz-mhmd: Unfortunately, a stabilization scenario may fail and throw exception if delay was not 0 as well. I updated it.

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Commons.java
moataz-mhmd: Change no longer needed.

aws-rds-dbclusterparametergroup/src/test/java/software/amazon/rds/dbclusterparametergroup/CreateHandlerTest.java
moataz-mhmd: done.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
```diff
@@ -304,53 +294,12 @@ private ProgressEvent<ResourceModel, CallbackContext> modifyParameters(final Pro
...
-                        3,
-                        () -> isDBClustersAvailable(proxyInvocation, model)))
-                .handleError((describeDbParameterGroupsRequest, exception, client, resourceModel, ctx) ->
-                        Commons.handleException(
-                                ProgressEvent.progress(resourceModel, ctx),
-                                exception,
-                                DB_CLUSTERS_STABILIZATION_ERROR_RULE_SET))
-                .progress();
+        CallbackContext callbackContext = progress.getCallbackContext();
+        if(!callbackContext.isParameterGroupStabilized())
```
Outige, line 300: You don't update this variable to indicate a successful stabilization. Could this lead to the lambda never getting stuck in a loop of stabilization? Or is this parameter updated implicitly?

aws-rds-dbclusterparametergroup/src/test/java/software/amazon/rds/dbclusterparametergroup/UpdateHandlerTest.java
```diff
@@ -159,77 +160,4 @@ public void handleRequest_Success() {
...
-        assertThat(handlerResponse.getErrorCode()).isNull();
-
-        verify(proxyRdsClient.client()).resetDBClusterParameterGroup(any(ResetDbClusterParameterGroupRequest.class));
-        verify(proxyRdsClient.client()).describeDBClusterParameters(any(DescribeDbClusterParametersRequest.class));
-        verify(proxyRdsClient.client(), times(2)).describeDBClustersPaginator(any(DescribeDbClustersRequest.class));
-        verify(proxyRdsClient.client()).describeDBClusterParameterGroups(any(DescribeDbClusterParameterGroupsRequest.class));
-        verify(proxyRdsClient.client()).listTagsForResource(any(ListTagsForResourceRequest.class));
-        verify(proxyRdsClient.client(), times(1)).addTagsToResource(any(AddTagsToResourceRequest.class));
-        verify(proxyRdsClient.client(), times(1)).removeTagsFromResource(any(RemoveTagsFromResourceRequest.class));
-    }
```
Outige, line 234: nit: Do we not want to add a test that explicitly tests that the code hangs for 5 minutes? Especially, so since we think this stabilization time might be subject to change, so it would be nice to just be able to update the test as a sanity check.

Outige (dismissed):

Outige (dismissed):

aws-rds-dbclusterparametergroup/src/test/java/software/amazon/rds/dbclusterparametergroup/UpdateHandlerTest.java
moataz-mhmd: Done.

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
```diff
@@ -19,4 +19,8 @@ public class HandlerConfig {
             .delay(Duration.ofSeconds(30))
             .timeout(Duration.ofMinutes(90))
             .build();
+
+    @Getter
+    @Builder.Default
+    final private int stabilizationInSeconds = 60;
```
osdrv, line 25: This looks foreign to the common library. I believe we need a different level of indirection to address this point. As far as I understand, the major reason to have it in the common config is to skip it in the tests. This could be achieved in multiple ways:
- A resource can introduce a local override of HandlerConfig and add the fields that are required.
- The `sleep/1` can be called conditionally, we can use `execOnce` guard for this purpose. In this case the timeout can be hardcoded.
- This config can be a property of the corresponding handler class. It should be pretty trivial to modify it upon a class construction, especially in the tests.

Frankly, I would prefer 1 of these approaches to getting the resource-specific config in this common class as it opens up a Pandora's box and ultimately accumulates random attributes and becomes a dusty corner. Let's avoid it now.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
```diff
@@ -51,7 +50,7 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
     protected static final String STACK_NAME = "rds";
     protected static final int MAX_LENGTH_GROUP_NAME = 255;
     // 5 min for waiting propagation according to https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyDBClusterParameterGroup.html
-    protected static final int NO_CALLBACK_DELAY = 0;
+    protected static final int STABILIZATION_DELAY_SECONDS = 5 * 60;
```
osdrv, line 53: The fact that we have to explain what time unit is being used in the var name makes me think we are using a wrong container (`int`). This is an internal property and it won't be serialized/deserialized. Let's use a more semantically-matching class `Duration` here. It could look like:
```java
protected static final Duration STABILIZATION_DELAY = Duration.ofMinutes(5);
```

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
```diff
@@ -304,53 +294,14 @@ private ProgressEvent<ResourceModel, CallbackContext> modifyParameters(final Pro
...
 
-    protected boolean isDBClustersAvailable(final ProxyClient<RdsClient> proxyClient, final ResourceModel model) {
-        return proxyClient
-                .injectCredentialsAndInvokeIterableV2(Translator.describeDbClustersRequest(), proxyClient.client()::describeDBClustersPaginator)
-                .stream()
-                .flatMap(describeDbClustersResponse -> describeDbClustersResponse.dbClusters().stream())
-                .filter(dbCluster -> dbCluster.dbClusterParameterGroup().equalsIgnoreCase(model.getDBClusterParameterGroupName()))
-                .allMatch(dbCluster -> dbCluster.status().equals(AVAILABLE));
-    }
+    protected ProgressEvent<ResourceModel, CallbackContext> waitForStabilization(final ProgressEvent<ResourceModel, CallbackContext> progress) {
```
osdrv, line 297: Let's call it `sleep`:
```java
protected ProgressEvent<ResourceModel, CallbackContext> sleep(final ProgressEvent<ResourceModel, CallbackContext> progress, final Duration delay) {
    return ProgressEvent.defaultInProgressHandler(progress.getCallbackContext(), delay.getSeconds(), progress.getResourceModel());
}
```

Looking at this method, it makes me think it could go to Commons.handler and take its well-deserved spot there.

Next, let's move the conditional execution out of the method: what we got now is `execIfNot(...)`, and we are trying to flatten this out to `if (not()) { exec(); }`. Note that this conditional execution could be guarded by `Commons.execOnce` (see DBInstance and DBCluster code for the ref). This could be 1 way to address the comment I left above (concerning adding the new attribute to the `HandlerConfig` class).

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/CallbackContext.java
```diff
@@ -16,8 +16,9 @@ public class CallbackContext extends StdCallbackContext implements TaggingContex
     private String dbClusterParameterGroupArn;
 
     private boolean parametersApplied;
-    private boolean clusterStabilized;
+    private boolean parameterGroupStabilized;
```
osdrv, line 19: Nit: just call it `stabilized`: it is clear from this context that we're talking about the owning resource.

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/HandlerConfig.java
moataz-mhmd: Thanks Oleg! Went with first choose to implement DBClusterParameterGroup of implementing local override of HandlerConfig.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
moataz-mhmd: `Commons.execOnce` relay on `then` to change `callbackContext` flag.
`Commons.execOnce` does not work with progress that have delay because `then` only applied when the delay is 0.
Here what happened 
1.  excOnce called with inProgressDelay progress
2. `then` did not called.
3. Flag did not updated.
4. CFN enters same step again after delay time is spent.
5. code enters `execOnce`again as flag was not updated
6. endless Loop.

I proposed solution to update `execOnce` implementation to 
```
    public static <M, C> ProgressEvent<M, C> execOnce(
            final ProgressEvent<M, C> progress,
            final ProgressEventLambda<M, C> func,
            final Function<C, Boolean> conditionGetter,
            final VoidBiFunction<C, Boolean> conditionSetter
    ) {
        if (!conditionGetter.apply(progress.getCallbackContext())) {
            ProgressEvent<M, C> progressEvent =  func.enact();
            conditionSetter.apply(progressEvent.getCallbackContext(), true);
        }
        return progress;
    }
```
But @khebul  had deep concern that will not work in all execOnce scenarios specifically stabilization scenario.  [here](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/292#discussion_r957604547) 

Here underlying implantation for more understanding
```
    public static <M, C> ProgressEvent<M, C> execOnce(
            final ProgressEvent<M, C> progress,
            final ProgressEventLambda<M, C> func,
            final Function<C, Boolean> conditionGetter,
            final VoidBiFunction<C, Boolean> conditionSetter
    ) {
        if (!conditionGetter.apply(progress.getCallbackContext())) {
            return func.enact().then(p -> {
                conditionSetter.apply(p.getCallbackContext(), true);
                return p;
            });
        }
        return progress;
    }
    //Underlying implemetation
    public ProgressEvent<ResourceT, CallbackT> then(Function<ProgressEvent<ResourceT, CallbackT>, ProgressEvent<ResourceT, CallbackT>> func) {
        return this.canContinueProgress() ? (ProgressEvent)func.apply(this) : this;
    }

    @JsonIgnore
    public boolean canContinueProgress() {
        return this.status == OperationStatus.IN_PROGRESS && this.callbackDelaySeconds == 0;
    }
```

osdrv (approved):

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/DefaultHandlerConfig.java
```diff
@@ -0,0 +1,16 @@
+package software.amazon.rds.dbclusterparametergroup;
+
+import java.time.Duration;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.experimental.SuperBuilder;
+
+@SuperBuilder
+public class DefaultHandlerConfig extends software.amazon.rds.common.handler.HandlerConfig {
```
khebul, line 10: nit: `DefaultHandlerConfig` implies that there are non-default handler configs. This class seems like a config for `DBClusterParameterGroup` handler.

osdrv: Proposed naming:
- `DBClusterParameterGroupHandlerConfig` (clear but unreasonably long)
- `HandlerConfigExt` (unclear, but short and efective)
- `LocalHandlerConfig` (optimal length-wise, but a bit unclear how far the zone of the locality spreads, though)

Outige (dismissed):

Outige (approved):

osdrv (approved):

PR #300, opened 2022-09-02
[Common] Move `test_handleRequest_error` to common

osdrv:
This commit extracts `test_handleRequest_error` introduced earlier in DBInstance and generalizes it to the level of being applicable cross-resource.


moataz-mhmd (approved):

aws-rds-cfn-common/pom.xml
```diff
@@ -72,7 +72,7 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <version>4.3.1</version>
-            <scope>test</scope>
```
khebul, line 75: Do we need mockito in runtime?

osdrv: This is the downside of keeping Test utils in cfn-common: any test utility dependency bubbles up into a non-test classpath. An alternative to that would be to split cfn-common-core and cfn-common-test.

khebul (approved):

PR #302, opened 2022-09-06
[DBClusterPG] DBParameterGroup parameter grouping method coppied for DBClusterPG

Outige:
This pr duplicate the logic used by DBParameterGroup to group dependent parameters for DBClusterParameterGroup.


khebul:
PR #303 was based on this request and merged instead

PR #311, opened 2022-09-16
[DBCluster] Enable In-place DBInstanceParameterGroup upgrade

khebul:
This CR adds ability to perform in-place DBInstanceParameterGroup update, making it possible to upgrade Major Engine Version of the clusters which members have custom DBParameterGroup set. 


aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -216,17 +219,27 @@ protected SecurityGroup fetchSecurityGroup(
         return response.securityGroups().get(0);
     }
 
+    protected DBInstance fetchDBInstance(
+            final ProxyClient<RdsClient> proxyClient,
+            final String dbInstanceIdentifier
+    ) {
+        final DescribeDbInstancesResponse response = proxyClient.injectCredentialsAndInvokeV2(
+                Translator.describeDbInstancesRequest(dbInstanceIdentifier),
+                proxyClient.client()::describeDBInstances
```
osdrv, line 228: Could you please check the corresponding FAS policies request this permission?

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -458,8 +471,17 @@ protected ProgressEvent<ResourceModel, CallbackContext> setDefaultVpcSecurityGro
         }
     }
 
-    protected boolean shouldSetDefaultVpcSecurityGroupIds(final ProgressEvent<ResourceModel, CallbackContext> progress) {
-        final ResourceModel resourceModel = progress.getResourceModel();
-        return CollectionUtils.isEmpty(resourceModel.getVpcSecurityGroupIds());
+    protected boolean shouldSetDefaultVpcSecurityGroupIds(final ResourceModel previousState, final ResourceModel desiredState) {
+        if (previousState != null) {
+            final List<String> previousVpcIds = CollectionUtils.isEmpty(previousState.getVpcSecurityGroupIds()) ?
+                    ImmutableList.of() : previousState.getVpcSecurityGroupIds();
```
osdrv, line 477: NIT: This is a variadic function returning a "real" list instance. Please prefer `Collections.emptyList()`.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/CallbackContext.java
```diff
@@ -3,19 +3,38 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
 @lombok.Getter
 @lombok.Setter
 @lombok.ToString
 @lombok.EqualsAndHashCode(callSuper = true)
 public class CallbackContext extends StdCallbackContext implements TaggingContext.Provider {
     private boolean modified;
+    private boolean isRebooted;
```
osdrv, line 15: NIT: please avoid modal verbs in var names here: a Boolean variable would trigger lombook to generate the corresponding `isRebooted` method.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/Translator.java
```diff
@@ -418,10 +424,18 @@ public static ResourceModel translateDbClusterFromSdk(
                 .build();
     }
 
-    public static DescribeGlobalClustersRequest describeGlobalClustersRequest(String globalClusterIdentifier) {
+    public static DescribeGlobalClustersRequest describeGlobalClustersRequest(final String globalClusterIdentifier) {
```
osdrv, line 427: Thx!

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/Translator.java
```diff
@@ -418,10 +424,18 @@ public static ResourceModel translateDbClusterFromSdk(
...
         return DescribeGlobalClustersRequest.builder().globalClusterIdentifier(globalClusterIdentifier).build();
     }
 
+    public static DescribeDbInstancesRequest describeDbInstancesRequest(final String dbInstanceIdentifier) {
+        return DescribeDbInstancesRequest.builder().dbInstanceIdentifier(dbInstanceIdentifier).build();
+    }
+
+    public static RebootDbInstanceRequest rebootDbInstanceRequest(final String dbInstanceIdentifier) {
```
osdrv, line 435: Sounds like we need to add this one to our FAS policy list as well

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/CallbackContext.java
```diff
@@ -3,19 +3,38 @@
...
+    public int getProbes(final String sampleName) {
+        return this.probes.getOrDefault(sampleName, 0);
+    }
+
+    public int incProbes(final String sampleName) {
+        return this.probes.merge(sampleName, 1, Integer::sum);
+    }
+
+    public void flushProbes(final String sampleName) {
+        this.probes.remove(sampleName);
```
osdrv, line 37: This bit of code has lasted much longer than I anticipated

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+        }
+
+        ProgressEvent<ResourceModel, CallbackContext> result = progress;
+
+        for (final DBClusterMember dbClusterMember: cluster.dbClusterMembers()) {
+            final ProgressEvent<ResourceModel, CallbackContext> p = proxy.initiate("rds::reboot-dbcluster-member-" + dbClusterMember.dbInstanceIdentifier(),
+                            proxyClient,
+                            progress.getResourceModel(),
+                            progress.getCallbackContext())
+                    .translateToServiceRequest(model -> Translator.rebootDbInstanceRequest(dbClusterMember.dbInstanceIdentifier()))
```
osdrv, line 163: This bit is super scary tbh. Do we reboot all the instances in the cluster unconditionally?

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+
+            if (p.isFailed() || p.isInProgressCallbackDelay()) {
+                return p;
+            }
+            result = p;
+        }
+        return result;
+    }
+
+    private boolean shouldRebootCluster(final ResourceModel previousResourceState, final ResourceModel desiredResourceState, boolean isRollback) {
```
osdrv, line 186: NIT: `final`

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+                !Objects.equals(previousResourceState.getDBInstanceParameterGroupName(), desiredResourceState.getDBInstanceParameterGroupName());
+    }
+
+    private boolean isDBInstanceStabilized(final ProxyClient<RdsClient> client, final String dbInstanceIdentifier, final ResourceModel resourceModel) {
+        final DBInstance dbInstance = fetchDBInstance(client, dbInstanceIdentifier);
+        final DBCluster dbCluster = fetchDBCluster(client, resourceModel);
+
+        return "available".equals(dbInstance.dbInstanceStatus()) &&
+                isParameterGroupApplied(dbInstance) &&
+                isClusterParameterGroupApplied(dbCluster, dbInstance);
```
osdrv, line 198: Is it sufficient? I believe DBInstance has a significantly more complicated definition of availability nowadays.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+            context.flushProbes(probeName);
+            return false;
+        }
+        context.incProbes(probeName);
+        if (context.getProbes(probeName) >= nProbes) {
+            context.flushProbes(probeName);
+            return true;
+        }
+        return false;
+    }
```
osdrv, line 232: As this is a travelling bit of code, let's move it to BaseHandler: contextually it is a generic helper

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+                            DEFAULT_DB_CLUSTER_ERROR_RULE_SET
+                    ))
+                    .progress();
+
+            if (p.isFailed() || p.isInProgressCallbackDelay()) {
+                return p;
+            }
+            result = p;
+        }
+        return result;
```
osdrv, line 183: Not sure I understand chaining in this case. `result` is always set to the newly initiated proxy invocation on line 159. Shall we chain them up with `then()` invocations?

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: TBF, we probably want to move a lot of the code in this PR to a common place. E.g. there is a lot of duplicated code between DBInstance and DBCluster. I think we want to do that as a separate PR.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: There is one extra stabilization condition for rebooting DBInstance: option group being in sync. For cluster upgrade it is not relevant.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: We do not do that unconditionally. `shouldReboot` method is checking for in-place DBInstanceParameterGoup update. For the specific reasons why we want to reboot at all - I am happy to discuss this verbally :)

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: It is possible to do that through chaining, yes. The current logic is that this function reboots one instance at the time: stabilization code returns in-progress `ProgressEvent` with correct callback context. This context is used in subsequent invocations.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: I think we might want to address the underlying reason for having this bit of code in the first place, but that is a bigger project.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+    }
+
+    private boolean shouldRebootCluster(final ResourceModel previousResourceState, final ResourceModel desiredResourceState, boolean isRollback) {
+        return !isRollback &&
```
moataz-mhmd, line 187: What will happen if failure happen in stack (other resource failure i.e: s3) after successfully updating DBCluster. So rollback will be initiated to revert   `DBInstanceParameterGroupName` but it will require reboot again to get instance available but this condition will stop it.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+                !Objects.equals(previousResourceState.getEngineVersion(), desiredResourceState.getEngineVersion()) &&
+                !Objects.equals(previousResourceState.getDBInstanceParameterGroupName(), desiredResourceState.getDBInstanceParameterGroupName());
+    }
+
+    private boolean isDBInstanceStabilized(final ProxyClient<RdsClient> client, final String dbInstanceIdentifier, final ResourceModel resourceModel) {
+        final DBInstance dbInstance = fetchDBInstance(client, dbInstanceIdentifier);
+        final DBCluster dbCluster = fetchDBCluster(client, resourceModel);
+
+        return "available".equals(dbInstance.dbInstanceStatus()) &&
```
moataz-mhmd, line 196: I know 2 more attributes that needed to be checked as well in DBInstance in case of update `DBInstanceClass, EngineVersion` as we saw failures before due to limited infrastructure capacity in region which result to modification is not applied although the API call succeeded and DBInstance returned to available status again.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -119,4 +139,95 @@ private boolean shouldRemoveFromGlobalCluster(
...
+        return result;
+    }
+
+    private boolean shouldRebootCluster(final ResourceModel previousResourceState, final ResourceModel desiredResourceState, boolean isRollback) {
+        return !isRollback &&
+                !Objects.equals(previousResourceState.getEngineVersion(), desiredResourceState.getEngineVersion()) &&
+                !Objects.equals(previousResourceState.getDBInstanceParameterGroupName(), desiredResourceState.getDBInstanceParameterGroupName());
+    }
+
+    private boolean isDBInstanceStabilized(final ProxyClient<RdsClient> client, final String dbInstanceIdentifier, final ResourceModel resourceModel) {
```
moataz-mhmd, line 192: I believe we should short circuit by throwing an exception if instance in available status but  `ParameterGroup` is not applied as we do not need to keep calling till stabilization timeout. There is nothing going to change if instance returned to available status but with ParameterGroup not applied. Same applies on any attributes we want to check.

aws-rds-dbcluster/docs/README.md
```diff
@@ -218,6 +220,16 @@ _Type_: String
...
 
+#### DBInstanceParameterGroupName
+
+The name of the DB parameter group to apply to all instances of the DB cluster.
+
+_Required_: No
+
+_Type_: String
+
+_Update requires_: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
```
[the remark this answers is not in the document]
khebul: That is automatically generated documentation which is entirely based on schema. The reboot is also staggered, so the cluster should stay available while it is performed.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: We do not support downgrades. Change in translator is needed to prevent setting the attribute on rollback.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: Not waiting on application of PG/CPG might trigger a stack rollback if the instance was also modified. I believe the safest thing to do is to wait for DBInstance availability. For full reasons why I am available for a verbal discussion, I think both you and Oleg might be interested.

aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
khebul: Hence, like in one of the comments I think that this kind of stabilization logic should be moved to a common place. But I belive that would be better done in a separate PR.

moataz-mhmd (dismissed):

Outige (dismissed):

osdrv (dismissed):

moataz-mhmd (dismissed):

Outige (dismissed):

osdrv (approved):

aws-rds-dbcluster/aws-rds-dbcluster.json
```diff
@@ -379,6 +383,7 @@
     "/properties/DBClusterIdentifier"
   ],
   "writeOnlyProperties": [
+    "/properties/DBInstanceParameterGroupName",
```
moataz-mhmd, line 386: AFAIU, Write only parameters will not return in read handler which means customer will not see the drift of `DBInstanceParameterGroupName` when he run `Detect Drift` action on stack although template value does not match physical value. Should we support it in Read handler or there is strong reason behind it?

moataz-mhmd (approved):

aws-rds-dbcluster/aws-rds-dbcluster.json
khebul: This is an intended behaviour. It is not possible to read this parameter from DBCluster, it is only meant to be supplied during Major Version Upgrade. Therefore, it is made a write-only parameter.

moataz-mhmd: But we can retrieve it from member DBInstance. If is upgraded manually out of band but stack still have old value. Customer run `Detect drift`  and it doesn't appear but when customer execute stack it will fail when trying to apply this parameter.
This case can happen as we did not support this feature before and customers might have scripts that do this manually.

PR #308, opened 2022-09-12
[Common] Split AbstractTestBase into TestHelper and TestCommon

khebul:
Split static functionality into TestHelper and TestCommon classes. In addition, this PR fixes compilation warning introduced by the earlier pull request.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #314, opened 2022-09-19
[Common] Split cfn-common and cfn-test-common

osdrv:
This commit splits up `aws-rds-cfn-common` package in 2.

The test helper suite is moved to a new package called `aws-rds-cfn-test-common`. The motivation for this change is to break down the common dependencies into compile and test scopes. The old approach forced us to include test dependencies into the compile scope, it is eliminated with this commit.

This PR superseeds https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/308.

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -26,6 +26,14 @@ jobs:
       run: pip install pre-commit cloudformation-cli cloudformation-cli-java-plugin
     - name: Run pre-commit
       run: pre-commit run --all-files
+    - name: Verify AWS::RDS::Test::Common
+      run: |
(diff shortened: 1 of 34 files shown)


.github/workflows/maven-verify.yml
```diff
@@ -26,6 +26,14 @@ jobs:
       run: pip install pre-commit cloudformation-cli cloudformation-cli-java-plugin
     - name: Run pre-commit
       run: pre-commit run --all-files
+    - name: Verify AWS::RDS::Test::Common
```
khebul, line 29: two action do exactly the same thing. Is it intended?

khebul (approved):

moataz-mhmd (approved):

Outige (approved):

PR #315, opened 2022-09-20
[TestCommon] Implement access permission verifier

osdrv:
This commit adds a mechanism to programmatically verify the schema permission integrity. The motivation for this change is to close the gap between the schema definition and the actual implementation. As practice indicates, one of the most common problems occurs when a new API call is being added and no corresponding access permission is requested by the schema.

The approach in a nutshell is to inspect the corresponding client object after every unit test execution and verify that all invocations were declared in the corresponding schema access permissions.

The public API for access permission verification is:

    verifyFASPermissions(clientMock);

It is supposed to be called upon every unit test completion, e.g:

    @AfterEach
    public void tearDown() {
        verifyFASPermissions(clientMock);
    }

The verifier will throw a MockitoAssertionError if the permission is missing.

### DIFF
--- a/aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/annotations/ExcludeFromJacocoGeneratedReport.java
+++ b/aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/annotations/ExcludeFromJacocoGeneratedReport.java
@@ -6,5 +6,5 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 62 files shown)


moataz-mhmd (commented):
Thanks a lot Oleg for implementing this. Once schema is updated, command `cfn generate` need to be executed to update `resource-roles.yml`. if this command is not run, Roles file will not be updated and cfn will know  nothing about new added permission and we are vunarble again to missing permission issue.
Can we force check against `cfn generate` command during PR verification so  we do not let missing role passes throw code change?

khebul (approved):

Outige (approved):

PR #322, opened 2022-09-28
[DBInstance] Move status definitions to enums

osdrv:
This commits performs a shallow refactoring of the status tracking layer which was implemented as a trivial set of string constants. `DBInstanceStatus` was the first case where we had to attach some additional meta-information to the status (terminal status flag).

Given our ultimate aim to let the stack expose unrecoverable problems as soon as possible, we are adding the same bit of information (terminal status flag) to `OptionGroupStatus`. This move motivates an introduction of a new inheritance chain.

Firstly, we introduce a `Status` interface which defines a minimalistic API for status-like enums. Whereas enums is a great way of handling discrete values, java adds some limitations on using the actual inheritance on these classes, hence we favor interfaces.

The second family grid or status-like objects is `TerminableStatus`. This interface declares a basic API to check whether this state is transient or terminal.

Other constant-defined statuses were moved to the same package and gained the same implementation of the base methods. Using this enum package is encouraged instead of const-based status handling.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -54,6 +54,7 @@ import software.amazon.awssdk.services.rds.model.InvalidRestoreException;
 import software.amazon.awssdk.services.rds.model.InvalidSubnetException;
 import software.amazon.awssdk.services.rds.model.InvalidVpcNetworkStateException;
 import software.amazon.awssdk.services.rds.model.KmsKeyNotAccessibleException;
+import software.amazon.awssdk.services.rds.model.OptionGroupMembership;
 import software.amazon.awssdk.services.rds.model.OptionGroupNotFoundException;
 import software.amazon.awssdk.services.rds.model.PendingModifiedValues;
 import software.amazon.awssdk.services.rds.model.ProvisionedIopsNotAvailableInAzException;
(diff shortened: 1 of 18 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -96,15 +103,7 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
...
-    static final String DB_PARAMETER_GROUP_STATUS_APPLYING = "applying";
-    static final String IN_SYNC_STATUS = "in-sync";
-    static final String PENDING_REBOOT_STATUS = "pending-reboot";
-    static final String READ_REPLICA_STATUS = "read replication";
-    static final String STORAGE_FULL_STATUS = "storage-full";
-    static final String READ_REPLICA_STATUS_REPLICATING = "replicating";
-    static final String VPC_SECURITY_GROUP_STATUS_ACTIVE = "active";
-    static final String DOMAIN_MEMBERSHIP_JOINED = "joined";
-    static final String DOMAIN_MEMBERSHIP_KERBEROS_ENABLED = "kerberos-enabled";
+    static final String READ_REPLICA_STATUS_TYPE = "read replication";
```
moataz-mhmd, line 106: That looks neat 👍

moataz-mhmd (approved):
Can we enhance this and add concept of transient state. A transient state is a state will disappear automatically without doing any modification after while i.e: `back-up` , `modifying`, `syncing`
And whenever meet these states, Before/After invoking action like delete,  modify, or create. We wait for it to disappear to continue our execution. 
We saw this need during  modifying and deleting scenrios.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/status/DBInstanceStatus.java
```diff
@@ -1,16 +1,17 @@
...
     Available("available"),
     Creating("creating"),
     Deleting("deleting"),
     Failed("failed", true),
     IncompatibleRestore("incompatible-restore", true),
     IncompatibleNetwork("incompatible-network", true),
     IncompatibleParameters("incompatible-parameters", true),
-    InaccessibleEncryptionCredentials("inaccessible-encryption-credentials", true);
+    InaccessibleEncryptionCredentials("inaccessible-encryption-credentials", true),
+    StorageFull("storage-full", true);
```
khebul, line 14: Did you test storage-full recovery with this code? My concern if describe during storage-full modify call returns a stale status, the whole execution fails.

osdrv: I see your point. `isDBInstanceStabilizedAfterMutate` relies on this status as a non-terminal. We'd better get rid of this flag here. Thanks.

moataz-mhmd (approved):

khebul (approved):

PR #323, opened 2022-09-29
[DBInstance] Cover more RDS Custom update paths

khebul:
This PR addresses oversights in Update logic that prevent some RDS Custom DBInstances from updating. API validation logic is very strict on which parameter the customer is allowed to include, so majority parameters need to be omitted if they are not changed between previous and desired model.

### DIFF
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/util/TranslatorUtils.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/util/TranslatorUtils.java
@@ -1,36 +1,32 @@
 package software.amazon.rds.common.util;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
+import software.amazon.awssdk.core.util.DefaultSdkAutoConstructList;
+import software.amazon.awssdk.core.util.DefaultSdkAutoConstructMap;
 
-import java.lang.reflect.Method;
-import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
 import java.util.Objects;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
 
 public class TranslatorUtils {
 
-    public static<M> M difference(final M previousModel, final M desiredModel, Supplier<M> instanceSupplier) {
-        if (Objects.isNull(previousModel)) {
-            return desiredModel;
+    public static <E, T extends List<E>> List<E> difference(final T prev, final T upd) {
+        if ((prev == null && upd == null) || (Objects.deepEquals(prev, upd))) {
+            return DefaultSdkAutoConstructList.getInstance();
         }
+        return upd;
+    }
 
-        try {
-            M result = instanceSupplier.get();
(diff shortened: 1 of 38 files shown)


aws-rds-cfn-common/src/main/java/software/amazon/rds/common/util/TranslatorUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class TranslatorUtils {
+
+    public static<M> M difference(final M previousModel, final M desiredModel, Supplier<M> instanceSupplier) {
```
moataz-mhmd, line 13: Is there any good library that can provide us same functionality?
I believe this could be popular issue. I would prefer to use library as reflection bugs are tough to discover and investigate. Also using good library could help here in future as this logic will be a good candidate to be used in post resource stabilization comparison  and logging resource differences as well.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -306,45 +306,46 @@ public static ModifyDbInstanceRequest modifyDbInstanceRequest(
...
+                .copyTagsToSnapshot(TranslatorUtils.difference(previousModel.getCopyTagsToSnapshot(), desiredModel.getCopyTagsToSnapshot()))
+                .dbInstanceClass(TranslatorUtils.difference(previousModel.getDBInstanceClass(), desiredModel.getDBInstanceClass()))
+                .dbInstanceIdentifier(TranslatorUtils.difference(previousModel.getDBInstanceIdentifier(), desiredModel.getDBInstanceIdentifier()))
+                .dbParameterGroupName(TranslatorUtils.difference(previousModel.getDBParameterGroupName(), desiredModel.getDBParameterGroupName()))
+                .dbPortNumber(translatePortToSdk(TranslatorUtils.difference(previousModel.getPort(), desiredModel.getPort())))
+                .deletionProtection(TranslatorUtils.difference(previousModel.getDeletionProtection(), desiredModel.getDeletionProtection()))
+                .domain(TranslatorUtils.difference(previousModel.getDomain(), desiredModel.getDomain()))
+                .domainIAMRoleName(TranslatorUtils.difference(previousModel.getDomainIAMRoleName(), desiredModel.getDomainIAMRoleName()))
```
osdrv, line 321: NIT: this still looks too verbose. I'd propose to change it to `DiffUtils.diff`. The name would rather be shorter: we are not adding any more clarity and only making the code less readable. Let's import this method statically: I believe in this case using a static import is justified for the sake of keeping it all-clear.

osdrv (approved):

moataz-mhmd (approved):

PR #327, opened 2022-10-05
[DBCluster] Stabilize on pending actions

### LINKED ISSUE #56: EnableIAMDatabaseAuthentication not applying to cluster
Not sure if this is the correct place to submit this.

Trying to set `IAM DB authentication` to true on the already existing cluster via cloudFormation, but when setting `EnableIAMDatabaseAuthentication: true` in the cloudFormation seems to have no effect on this. Am I reading the documentation correct? I was looking at this page [AWS::RDS::DBCluster](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbcluster.html#cfn-rds-dbcluster-enableiamdatabaseauthentication)

osdrv:
This commit adds an additional condition to the DBCluster stabilization logic: apart from checking that the cluster has achieved an Available state, it also checks there are no pending actions scheduled for an immediate execution.

The motivation for this change is the RDS API specifics: whereas some attributes are being applied immediately, the rest could be scheduled up for a delayed execution and picked up as soon as possible. This is the case for the following attributes:
  - `masterUserPassword`
  - `iamDatabaseAuthenticationEnabled`
  - `engineVersion`

This commit ensures the handler will inspect `pendingActions` before moving on.

Fixes #56.

### DIFF
--- a/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
+++ b/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
@@ -16,6 +16,7 @@ import software.amazon.awssdk.services.ec2.Ec2Client;
 import software.amazon.awssdk.services.ec2.model.DescribeSecurityGroupsResponse;
 import software.amazon.awssdk.services.ec2.model.SecurityGroup;
 import software.amazon.awssdk.services.rds.RdsClient;
+import software.amazon.awssdk.services.rds.model.ClusterPendingModifiedValues;
 import software.amazon.awssdk.services.rds.model.DBCluster;
 import software.amazon.awssdk.services.rds.model.DBSubnetGroup;
 import software.amazon.awssdk.services.rds.model.DbClusterAlreadyExistsException;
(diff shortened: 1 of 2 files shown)


aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -221,12 +222,25 @@ protected boolean isGlobalClusterMember(final ResourceModel model) {
         return StringUtils.isNotBlank(model.getGlobalClusterIdentifier());
     }
 
+    protected boolean isDBClusterAvailable(final DBCluster dbCluster) {
+        return DBClusterStatus.Available.equalsString(dbCluster.status());
+    }
+
+    protected boolean isNoPendingChanges(final DBCluster dbCluster) {
+        final ClusterPendingModifiedValues modifiedValues = dbCluster.pendingModifiedValues();
```
khebul, line 230: Maintenance actions are applied during maintenance. I believe these changes are applied in a few cases: 
* After [ApplyPendingMaintenanceAction] API call (https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ApplyPendingMaintenanceAction.html)
* During Maintenance Window
* After ModifyDBCluster call with --applyImmediately

Sometimes there is a short delay before pending maintenance action is picked by RDS after a modify call. In that case the stabilization logic could run forever (or rather up to 36 hours or until next maintenance). Can you confirm that this is not the case?

khebul (approved):

moataz-mhmd (approved):
Should we force applyingPendingActions if we found any. We do similar thing in DBInstance as we force changes that needs instance reboot to apply.

PR #328, opened 2022-10-06
[DBCluster] Only update difference in associated roles.

moataz-mhmd:
Change the unconditional way we update associated roles in DBCluster to only update the difference in roles between previous state and desired resource.

### DIFF
--- a/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
+++ b/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
@@ -80,8 +80,12 @@ public class UpdateHandler extends BaseHandlerStd {
                         () -> modifyDBCluster(proxy, rdsProxyClient, progress, previousResourceState, desiredResourceState, isRollback),
                         CallbackContext::isModified,
                         CallbackContext::setModified))
-                .then(progress -> removeAssociatedRoles(proxy, rdsProxyClient, progress, setDefaults(request.getPreviousResourceState()).getAssociatedRoles()))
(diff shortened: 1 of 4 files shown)


aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/UpdateHandler.java
```diff
@@ -80,8 +80,12 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
                         () -> modifyDBCluster(proxy, rdsProxyClient, progress, previousResourceState, desiredResourceState, isRollback),
                         CallbackContext::isModified,
                         CallbackContext::setModified))
-                .then(progress -> removeAssociatedRoles(proxy, rdsProxyClient, progress, setDefaults(request.getPreviousResourceState()).getAssociatedRoles()))
-                .then(progress -> addAssociatedRoles(proxy, rdsProxyClient, progress, progress.getResourceModel().getAssociatedRoles()))
+                .then(progress -> updateAssociatedRoles(
+                        proxy,
+                        rdsProxyClient,
+                        progress,
+                        setDefaults(request.getPreviousResourceState()).getAssociatedRoles(),
```
osdrv, line 87: I believe this mutating call could cause unexpected side effects here: the `setDefaults` has an effect on the entire previous resource state. Passing `request.getPreviousResourceState.getAssociatedRoles()` should be enough.

updateAssociatedRoles should be able to handle nullable collections. This would be enough.

moataz-mhmd: Done. Thanks Oleg!

osdrv (approved):

khebul (approved):

PR #331, opened 2022-10-11
[Common] Rework runtime config system

osdrv:
This commit introduces a new config system called `RuntimeConfig` as a replacement of `HandlerConfig`.

The motivation for the change is to move away from the hardcoded config to a local file-backed extendable system.

As the primary config data medium we chose plain old properties files. It is not the best way to store structured data, but works well enough for flat primitive-only configs.

A resource using RuntimeConfig is expected to expose a new config properties folder and add it to the compiled classpath.

The default config sets are:
- `config/runtime.properties`
- `config/test.properties`

### DIFF
--- a/aws-rds-dbinstance/config/runtime.properties
+++ b/aws-rds-dbinstance/config/runtime.properties
@@ -1,0 +1,2 @@
+handler.backoff.delay=30000
+handler.backoff.timeout=129600000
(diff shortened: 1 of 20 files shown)


aws-rds-dbinstance/config/runtime.properties
```diff
@@ -0,0 +1,2 @@
+handler.backoff.delay=30000
+handler.backoff.timeout=129600000
```
khebul, line 2: I believe this is quite hard to read, and unless you know this is in milliseconds. Can we utilize [ISO8601 durations](https://en.wikipedia.org/wiki/ISO_8601#Durations)?

moataz-mhmd (approved):

PR #343, opened 2022-10-25
[DBCluster] Set port conditionally on modify after create

osdrv:
This commit continues the conditional modify-db-cluster attribute set approach applied in DBInstance. In particular, setting `port` attribute is now conditional as RDS API is quite inconsistent in accepting the same-value parameters: some engines/engine modes would accept it just fine if it is a no-change set, but some would reject agressively.

### DIFF
--- a/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/CreateHandler.java
+++ b/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/CreateHandler.java
@@ -163,7 +163,7 @@ public class CreateHandler extends BaseHandlerStd {
             final ProgressEvent<ResourceModel, CallbackContext> progress
     ) {
         return proxy.initiate("rds::modify-dbcluster", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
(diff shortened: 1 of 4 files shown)


aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/CreateHandler.java
```diff
@@ -163,7 +163,7 @@ protected ProgressEvent<ResourceModel, CallbackContext> modifyDBCluster(
             final ProgressEvent<ResourceModel, CallbackContext> progress
     ) {
         return proxy.initiate("rds::modify-dbcluster", proxyClient, progress.getResourceModel(), progress.getCallbackContext())
-                .translateToServiceRequest(Translator::modifyDbClusterRequest)
+                .translateToServiceRequest(model -> Translator.modifyDbClusterRequest(model, model, false))
```
moataz-mhmd, line 166: should be `Translator.modifyDbClusterRequest(previousModel, desiredModel, false))` instead of `Translator.modifyDbClusterRequest(model, model, false))`

khebul (approved):

moataz-mhmd (approved):

PR #345, opened 2022-10-27
[DBInstance] Check events on stabilization

osdrv:
This commit introduces an additional check that would be performed upon a DBInstance stabilization. The motivation for this change is to extend a limited set of criterias in use to declare an operation completion and its status.

In particular, using events could be the only way to detect a mutation issue when RDS API indicates that the initial request returns a successful response and the resource eventually reaches an `available` status, despite the fact that the mutation failed and only left a corresponding trace in the event stream.

Using events in stabilization requires an additional permission: rds:DescribeEvents. Given this permission has never been a part of the DBInstance permission set, we wrap it in a silent try-catch block so to avoid turning events into a hard dependency (at least, for now).


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -544,19 +580,48 @@ private void assertNoTerminalStatus(final DBInstance dbInstance) throws CfnNotSt
...
+        if (isDBInstanceAvailable(dbInstance) &&
                 isReplicationComplete(dbInstance) &&
                 isDBParameterGroupNotApplying(dbInstance) &&
                 isNoPendingChanges(dbInstance) &&
                 isVpcSecurityGroupsActive(dbInstance) &&
                 isDomainMembershipsJoined(dbInstance) &&
-                isPromotionTierUpdated(dbInstance, model);
+                isPromotionTierUpdated(dbInstance, model)
+        ) {
+            assertNoFailedEvents(rdsProxyClient, model, context);
```
khebul, line 598: In theory, we could fail faster by checking events rather then performing the rest of stabilization. It might have more complex interaction with rollbacks, however, since instance might not be ready for modification if we fail fast.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/UpdateHandler.java
```diff
@@ -122,16 +123,18 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
                     }
                     return progress;
                 }, CallbackContext::isReadReplicaPromoted, CallbackContext::setReadReplicaPromoted))
-                .then(progress -> Commons.execOnce(progress, () ->
-                                versioned(proxy, rdsProxyClient, progress, null, ImmutableMap.of(
+                .then(progress -> Commons.execOnce(progress, () -> {
+                                    progress.getCallbackContext().timestampOnce(RESOURCE_UPDATED_AT, Instant.now());
```
khebul, line 127: before that point we could have sent up to 2 modify calls. Might be worth to set the variable earlier

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -448,6 +458,32 @@ protected boolean isRdsCustomOracleInstance(final ResourceModel model) {
...
     }
 
+    protected List<Event> fetchEvents(
+            final ProxyClient<RdsClient> rdsProxyClient,
+            final ResourceModel model,
+            final String eventType,
+            final Instant fetchSince
+    ) {
+        DescribeEventsResponse response = rdsProxyClient.injectCredentialsAndInvokeV2(
+                DescribeEventsRequest.builder()
```
khebul, line 468: I think we are missing `duration` parameter from the describe call. It defaults to 60 minutes. If any event happens after 60 minutes since the start of the update handler, we will fail to detect it.

Ideally, we would also want logical clocks instead of UTC ones, but alas RDS API does not provide that functionality :(

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -544,19 +580,48 @@ private void assertNoTerminalStatus(final DBInstance dbInstance) throws CfnNotSt
...
+    }
+
+    protected void assertNoFailedEvents(
+            final ProxyClient<RdsClient> rdsProxyClient,
+            final ResourceModel model,
+            final CallbackContext context
+    ) {
+        final Instant fetchSince = context.getTimestamp(RESOURCE_UPDATED_AT);
+        if (fetchSince == null) {
+            // we did not timestamp it before,
```
khebul, line 611: I think this has a chance of happening when sending Modify call to recover from storage-full scenario by invoking modify call with allocated-storage only. Should we then use named timestamps (i.e. `allocate-storage-start`)?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -544,19 +580,48 @@ private void assertNoTerminalStatus(final DBInstance dbInstance) throws CfnNotSt
...
+            return;
+        }
+        List<Event> notifications;
+        try {
+            notifications = fetchEvents(rdsProxyClient, model, "notification", fetchSince);
+        } catch (AwsServiceException e) {
+            // best effort
```
khebul, line 618: In the worst case we will default to current behavior. Customers will not see their stack failing, but rather would be able to discover the changes by runnng drift. I think SDK client makes a few attempts already before giving up, so the chance of it all happening during normal operation are rather slim. I think we can keep the best effort approach.

I think we should distinguish between different kinds of exceptions, though. For example, on throttling we might perform retry at a later point.

khebul (approved):

Outige (approved):

PR #357, opened 2022-11-16
[DBInstance] Modify with a lookback at the current state after create

osdrv:
This commit changes the way `CreateHandler` performs an update after create. Historically, not every attribute of DBInstance could be set with a `CreateDBInstance`/`RestoreDBInstanceFromDBSnapshot`/`CreateDBInstanceReadReplica` request. For this reason the handler would emit a follow-up `ModifyDBInstance` request that would ensure the right attribute set. Since `DifferenceUtils` was introduced, `Translator` started assembling minimalistic modify requests, only including the attributes that are about to be changed. Given that a `CreateHandler` request contains no previous resource state, we were using a barebone model builder and feed it to diff. This gives an all-changed modify request. This commit alters this approach and describes a DBInstance right before modifying it. Using this technique allows us to use `diff` effectively and emit minimalistic modify requests.

This commit embeds a series of smaller changes:
- Apart from oracle-custom, we're adding a set of sqlserver-custom engines and gather them under a custom engine family umbrella.
- `DeletionPolicy: Snapshot` is rejected upon a create for custom engine family.
- Final snapshot is skipped for custom engine family.
- Translator avoids creating empty collections on from-sdk and to-sdk conversions if a null input is provided. This allows us to skip sending meaningless empty collections in modifies.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -1,7 +1,6 @@
 package software.amazon.rds.dbinstance;
 
 import java.time.Duration;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
@@ -15,7 +14,6 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import com.google.common.collect.ImmutableList;
 import org.apache.commons.lang3.BooleanUtils;
 
 import com.amazonaws.util.CollectionUtils;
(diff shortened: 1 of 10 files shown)


khebul:
I think the general idea of only modifying required parameters is a correct one. However we have to be very careful in how we implement this. For example, customers are allowed to specify only Major engine version in their resource definition. DescribeDBInstance, on the other hand, always returns exact version. If this is not taken into account, we could cause unexpected instance reboots/unavailability. 

I am aware about the following inconsistencies:
* EngineVersion will be changed during maintenance if AutoMinorVersionUpgrade option is specified
* KMS Keys can be specified by either Name or ARN, while describe-db-instances API returns canonized version
* DBInstanceClass can be automatically upgraded on instance type retirement
* AWS reserved tags can be added by non-CFN controlled agents

And this is not an exhaustive list of parameters. I think we might do "diff onboarding" in a controlled way, and slowly expand the list over the course of migration process

moataz-mhmd (commented):
I think we will need to have the ability for injecting custom cooperators in Difference mechanism that is aware of nature of attribute i.e: 
For EgineVersion, Ability to compare and ignore downgrades. Understands different engines types and how their major.minor.

PR #360, opened 2022-11-18
[GlobalCluster] Added major version upgrade support to globalCluster …

feiyuren233:
[GlobalCluster]

[1179](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1179)

Added major version upgrade support to globalCluster 

Please note: this change doesn't support minor version upgrade, the usual method of manually upgrading all member clusters before patching primary still applies.

### DIFF
--- a/aws-rds-globalcluster/aws-rds-globalcluster.json
+++ b/aws-rds-globalcluster/aws-rds-globalcluster.json
@@ -59,8 +59,7 @@
     "/properties/GlobalClusterIdentifier",
     "/properties/SourceDBClusterIdentifier",
     "/properties/StorageEncrypted",
-    "/properties/Engine",
-    "/properties/EngineVersion"
+    "/properties/Engine"
   ],
   "primaryIdentifier": [
     "/properties/GlobalClusterIdentifier"
--- a/aws-rds-globalcluster/docs/README.md
+++ b/aws-rds-globalcluster/docs/README.md
@@ -58,7 +58,7 @@ _Required_: No
 
 _Type_: String
 
-_Update requires_: [Replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-replacement)
+_Update requires_: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
 
 #### DeletionProtection
 
--- a/aws-rds-globalcluster/src/main/java/software/amazon/rds/globalcluster/Translator.java
+++ b/aws-rds-globalcluster/src/main/java/software/amazon/rds/globalcluster/Translator.java
@@ -35,13 +35,17 @@ public class Translator {
             .build();
   }
 
-  static ModifyGlobalClusterRequest modifyGlobalClusterRequest(final ResourceModel previousModel, final ResourceModel desiredModel) {
+  static ModifyGlobalClusterRequest modifyGlobalClusterRequest(
+          final ResourceModel previousModel,
+          final ResourceModel desiredModel,
+          final boolean isRollback
+  ) {
     ModifyGlobalClusterRequest.Builder builder = ModifyGlobalClusterRequest.builder()
             .globalClusterIdentifier(desiredModel.getGlobalClusterIdentifier())
             .deletionProtection(desiredModel.getDeletionProtection());
 
     if (previousModel != null) {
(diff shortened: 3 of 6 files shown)


aws-rds-globalcluster/aws-rds-globalcluster.json
```diff
@@ -59,8 +59,7 @@
     "/properties/GlobalClusterIdentifier",
     "/properties/SourceDBClusterIdentifier",
     "/properties/StorageEncrypted",
-    "/properties/Engine",
-    "/properties/EngineVersion"
```
ammokhov, line 63: by removing this you are implying a significant change in resource behavior - now customer should not receive any interruptions. My questions is about what the experience should be. Engine version upgrade should cause no downtime but what about downgrade. Will customer be able to successfully provision lower version? If not then, we should think about preserving replacement behavior on version downgrade. One way of doing so is to mark `EngineVersion` as `conditionalCreateOnly`. This would imply that resource will decide whether to replace during the runtime

ammokhov (changes requested):

aws-rds-globalcluster/aws-rds-globalcluster.json
feiyuren233: Good point but i think this behaviour is consistent with regular cluster update requests, the modify cluster request should just fail without replacement.

aws-rds-globalcluster/docs/README.md
```diff
@@ -58,7 +58,7 @@ _Required_: No
 
 _Type_: String
 
-_Update requires_: [Replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-replacement)
+_Update requires_: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
```
Dmytrok-amz, line 61: That's actually not accurate - major version upgrade does cause interruptions, all of the global cluster goes down for the duration of MVU. "Updates with Some Interruption" from the link looks like a more accurate representation of the change.

feiyuren233: Good point, however, the "Updates with Some Interruption" flag doesn't appear to be used anywhere in this repo... not sure about the purpose of it in the doc...

aws-rds-globalcluster/aws-rds-globalcluster.json
ammokhov: so we are ok making this feature a forward only update?

aws-rds-globalcluster/src/main/java/software/amazon/rds/globalcluster/Translator.java
```diff
@@ -33,11 +35,19 @@ static CreateGlobalClusterRequest createGlobalClusterRequest(final ResourceModel
...
+    ModifyGlobalClusterRequest.Builder builder = ModifyGlobalClusterRequest.builder()
+            .globalClusterIdentifier(desiredModel.getGlobalClusterIdentifier())
+            .deletionProtection(desiredModel.getDeletionProtection());
+
+    if (previousModel != null) {
+      if (!Objects.equals(previousModel.getEngineVersion(), desiredModel.getEngineVersion())) {
+        builder.engineVersion(desiredModel.getEngineVersion());
+        builder.allowMajorVersionUpgrade(true);
+      }
+    }
```
ammokhov, line 48: I think you should incorporate `rollback` flag and if it is true and previous version is lower then current, then you should not include `EngineVersion` in the request, otherwise rollback will fail. Rollback is a revert update

aws-rds-globalcluster/aws-rds-globalcluster.json
feiyuren233: Correct me if I'm wrong but I think the engine version upgrade has been a forward-only thing for other resource types as well. This change just makes it consistent with other types.

Compared with DbCluster type: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/37268603fa31226859ed4cf3af9511bce4ae294d/aws-rds-dbcluster/aws-rds-dbcluster.json#L383-L402

https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/37268603fa31226859ed4cf3af9511bce4ae294d/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/Translator.java#L223-L226

ammokhov (approved):

Dmytrok-amz (approved):
LGTM. Unfortunate that we don't distinguish between "No interruption" and "Some interruption", but it is what it is

Issue #373, opened 2022-12-16
AWS::RDS::DBCluster AssociatedRoles translation error


evadne:
Given the following abridged template, given the `IsDatabaseImportExportEnabled` condition is true, the resource is created/updated with only 1 associated role for `s3Import` 

Reason Why This Is Wrong:
- The template I have created, expects the same IAM role to be added for both `s3Import` and `s3Export` features.
- The template executed via CFN without errors and there was no drift detected.
- However s3Export did not work.
- As the cluster is Amazon Aurora with PostgreSQL compatibility I have viewed documentation at https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/AuroraMySQL.Integrating.Authorizing.IAM.AddRoleToDBCluster.html.
- When I realised that the role was only added for `s3Import` feature, via checking the DB Cluster in the AWS Console, I attempted to add the same role for `s3Export` however the console rejected this and gave me an error message saying that the same role  has already been added.
- The document linked above did not mention that the same role can only be used for one feature.

Workaround:
- I have created a 2nd IAM role and associated it for the `s3Export` feature

```yaml
DatabaseCluster:
    Type: AWS::RDS::DBCluster
    DependsOn:
      - DatabaseClusterParameterGroup
      - DatabaseInstanceParameterGroup
    Properties:
      AssociatedRoles:
        - !If
          - IsDatabaseImportExportEnabled
          - FeatureName: s3Import
            RoleArn: !GetAtt DatabaseImportExportRole.Arn
          - !Ref AWS::NoValue
        - !If
          - IsDatabaseImportExportEnabled
          - FeatureName: s3Export
            RoleArn: !GetAtt DatabaseImportExportRole.Arn
          - !Ref AWS::NoValue
      BackupRetentionPeriod: !Ref DatabaseBackupRetentionPeriod
      DatabaseName: postgres
      DBClusterParameterGroupName: !Ref DatabaseClusterParameterGroup
      DBSubnetGroupName: !Ref DatabaseSubnetGroup
      EnableCloudwatchLogsExports:
        - postgresql
      Engine: aurora-postgresql
      EngineVersion: '12.12'
      MasterUsername: !Sub '{{resolve:secretsmanager:${DatabaseMasterUsernameSecret}:SecretString::}}'
      MasterUserPassword: !Sub '{{resolve:secretsmanager:${DatabaseMasterPasswordSecret}:SecretString::}}'
[...]

PR #375, opened 2022-12-19
[DBCluster] Report `DbClusterRoleAlreadyExistsException` exception

osdrv:
This commit alters the error rule set for adding associated roles to DBCluster. The current implementation soft-fails an already-exists exception for historical reasons.

The error can occur in 2 major cases:
  - Update failed to remove a role and a rollback would fail due to a resource conflict.
  - The same role was added under distinct feature names and this specific engine requires all roles to be unique.

This CR preserves soft-failing for a failed rollback (recovery mode) but exposes `DbClusterRoleAlreadyExistsException` on a regular update.

Refs https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/issues/373.

### DIFF
--- a/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
+++ b/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
@@ -102,6 +102,7 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
                     StorageQuotaExceededException.class,
                     SnapshotQuotaExceededException.class)
             .withErrorClasses(ErrorStatus.failWith(HandlerErrorCode.ResourceConflict),
+                    DbClusterRoleAlreadyExistsException.class,
                     InvalidDbClusterStateException.class,
                     InvalidDbInstanceStateException.class,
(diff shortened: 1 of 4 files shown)


aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -335,7 +338,8 @@ protected ProgressEvent<ResourceModel, CallbackContext> removeAssociatedRoles(
             final AmazonWebServicesClientProxy proxy,
             final ProxyClient<RdsClient> proxyClient,
             final ProgressEvent<ResourceModel, CallbackContext> progress,
-            final Collection<DBClusterRole> roles
+            final Collection<DBClusterRole> roles,
+            final boolean recoveryMode
```
khebul, line 342: nit: might be worth to keep this consistent with other use cases and name this "isRollback"

osdrv: Fixed. Thx.

khebul (approved):

Outige (approved):

PR #378, opened 2022-12-28
[DBInstance] Unset `AllocatedStorage` and `Iops` on rollback

osdrv:
This commit changes the handler behavior on a rollback update. The current strategy is to ensure `AllocatedStorage` and `Iops` changeability. Effectively, this means that the handler would ensure that neither of these attributes is being reduced upon a rollback. An attempt to reduce the corresponding attribute is known to be rejected by the RDS API, hence the handler would keep the old (pre-rollback) value. This commit alters this strategy to providing no pre-rollback values at all if either of these attributes is identifier as non-updatable. The translator would provide a null-value instead.

The motivation behind this change is to expand the applicability of the "keep whatever is working" on a case when the initial update failed. The existing approach would only work if the initial change was applied successfully and failure happened somewhere down the stack tree. The change would be invalid if the initial update never went through.

Refs https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1312.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
@@ -326,7 +326,6 @@ public class Translator {
                 .dbInstanceIdentifier(desiredModel.getDBInstanceIdentifier())
                 .dbParameterGroupName(diff(previousModel.getDBParameterGroupName(), desiredModel.getDBParameterGroupName()))
                 .dbSecurityGroups(diff(previousModel.getDBSecurityGroups(), desiredModel.getDBSecurityGroups()))
-                .engineVersion(diff(previousModel.getEngineVersion(), desiredModel.getEngineVersion()))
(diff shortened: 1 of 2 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -403,10 +396,10 @@ public static ModifyDbInstanceRequest modifyDbInstanceRequest(
 
         if (BooleanUtils.isTrue(isRollback)) {
             builder.allocatedStorage(
-                    canUpdateAllocatedStorage(previousModel.getAllocatedStorage(), desiredModel.getAllocatedStorage()) ? getAllocatedStorage(desiredModel) : getAllocatedStorage(previousModel)
```
khebul, line 406: Is it intentional that `modifyDbInstanceRequestV12` and `modifyDbInstanceRequest` use different conditions?

khebul (approved):

Outige (approved):

PR #381, opened 2023-01-10
[DBInstance] Certificate enhancements

khebul:
This CR adds a few enhancements to how DBInstace works with Certificates:
* Added ability to specify CA Certificate on create
* Added a new parameter that would allow applying CA rotation on next maintenance rather then rebooting DBInstance immediately
* Added read-only parameters that would allow to read CA Certificate CAIdentifier/ValidTill date


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -609,7 +611,7 @@ protected boolean isDBInstanceStabilizedAfterMutate(
         return isDBInstanceAvailable(dbInstance) &&
                 isReplicationComplete(dbInstance) &&
                 isDBParameterGroupNotApplying(dbInstance) &&
-                isNoPendingChanges(dbInstance) &&
+                isNoPendingChanges(dbInstance, model) &&
```
osdrv, line 614: No need to pack this up in a single method call, given how foreign this signature looks. Let's add a separate method to check model state. It will keep these predicates signature-consistent and provide a logical equivalent of this combo-check.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -753,6 +757,13 @@ public static List<String> translateVpcSecurityGroupsFromSdk(
                 .collect(Collectors.toList());
     }
 
+    public static CertificateDetails translateCertificateDetailsFromSdk(software.amazon.awssdk.services.rds.model.CertificateDetails certificateDetails) {
+        return certificateDetails == null ? null : CertificateDetails.builder()
+                .cAIdentifier(certificateDetails.caIdentifier())
+                .validTill(certificateDetails.validTill().toString()).
```
osdrv, line 763: Any chance `certificateDetails.validTill()` could return null?

osdrv (approved):

dbbh (approved):

PR #383, opened 2023-01-16
[Common] Implement Drift Detector Test Helper

osdrv:
This commit implements a new test helper called `ResourceDriftTestHelper`. This is an assertion helper that checks if there is a drift between an input and an output, based on the resource schema definition provided.

In particular, the drift detector would engage `propertyTransform` instructions from the schema and compare the transform formula result with the output. If neither the original input value nor any of the transformation outputs match the output, the resource is considered to have drifted.

### DIFF
--- a/aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/schema/ResourceDriftTestHelper.java
+++ b/aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/schema/ResourceDriftTestHelper.java
@@ -1,0 +1,139 @@
+package software.amazon.rds.test.common.schema;
+
+import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.assertj.core.api.Assertions;
+import org.json.JSONObject;
+
+import com.api.jsonata4java.expressions.EvaluateException;
+import com.api.jsonata4java.expressions.Expressions;
+import com.api.jsonata4java.expressions.ParseException;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableSet;
+
+public class ResourceDriftTestHelper {
+
+    private final static Set<Class<?>> primitiveTypes = ImmutableSet.of(
+            Integer.class,
+            String.class,
+            Boolean.class
+    );
+
+    private static final String PROPERTIES = "/properties";
+    private static final String PROPERTY_OR_SPLIT_REGEX = "\\s+\\$OR\\s+";
+    private static final String PROPERTY_PATH_SEPARATOR = "/";
+
+    public static void assertResourceNotDrifted(
+            final Object input,
+            final Object output,
+            final JSONObject resourceSchema
+    ) {
+        assertResourceNotDrifted(input, output, resourceSchema, PROPERTIES);
+    }
+
+    public static void assertResourceNotDrifted(
+            final Object input,
+            final Object output,
+            final JSONObject resourceSchema,
+            final String basePath
+    ) {
+        if (input == null || output == null) {
+            Assertions.assertThat(input).isEqualTo(output);
+            return;
+        }
+        Assertions.assertThat(input).hasSameClassAs(output);
+
+        final JSONObject propertyTransformMap = resourceSchema.getJSONObject("propertyTransform");
+        final ObjectMapper objectMapper = new ObjectMapper();
+
+        JsonNode inputJsonNode = null;
+        try {
+            // JSONata accepts JsonNode as an input. The easiest way to get a JsonNode from an object
+            // is to convert it to a string and parse immediately.
+            inputJsonNode = objectMapper.readTree(objectMapper.writeValueAsString(input));
+        } catch (JsonProcessingException e) {
+            Assertions.fail(e.getMessage());
+        }
+
+        final Field[] fields = input.getClass().getDeclaredFields();
+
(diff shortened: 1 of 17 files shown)


aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/schema/ResourceDriftTestHelper.java
```diff
@@ -0,0 +1,139 @@
...
+public class ResourceDriftTestHelper {
+
+    private final static Set<Class<?>> primitiveTypes = ImmutableSet.of(
+            Integer.class,
+            String.class,
+            Boolean.class
+    );
+
+    private static final String PROPERTIES = "/properties";
+    private static final String PROPERTY_OR_SPLIT_REGEX = "\\s*\\$OR\\s*";
```
khebul, line 30: is expression line `hello$ORworld` a valid expression? I am mainly asking due to `*` qualifier instead of `+`

aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/schema/ResourceDriftTestHelper.java
```diff
@@ -0,0 +1,139 @@
...
+            field.setAccessible(true);
+            Object inputFieldVal = null;
+            Object outputFieldVal = null;
+            try {
+                inputFieldVal = field.get(input);
+                outputFieldVal = field.get(output);
+            } catch (IllegalAccessException e) {
+                Assertions.fail(e.getMessage());
+            }
+            if (Objects.equals(inputFieldVal, outputFieldVal)) {
```
khebul, line 84: If the field is missing in `input`, but is present in `output` that is also not considered as drift. Exception to it is arrays.

aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/schema/ResourceDriftTestHelper.java
osdrv: Fixed with `+`

aws-rds-cfn-test-common/src/main/java/software/amazon/rds/test/common/schema/ResourceDriftTestHelper.java
osdrv: In this case i would prefer to keep the original logic in place. The purpose of this tool is not to completely emulate CFN behavior with all its conventions, but rather test the logic contained in `propertyTransform`. In this case we are risking of creating a false-positive check scenario when we've gotten a malformed input. Given these tests won't provide any specific value (i.e. it won't prove or disprove the logic), we could ignore this "relaxed" behavior of the drift detector.

aws-rds-dbinstance/aws-rds-dbinstance.json
```diff
@@ -447,7 +447,9 @@
     "/properties/OptionGroupName": "$lowercase(OptionGroupName)",
     "/properties/SourceDBInstanceAutomatedBackupsArn": "$lowercase(SourceDBInstanceAutomatedBackupsArn)",
     "/properties/SourceDBInstanceIdentifier": "$lowercase(SourceDBInstanceIdentifier)",
-    "/properties/MasterUserSecret/KmsKeyId": "$join([\"arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:kms:[a-z]{2}[-]{1}[a-z]{3,10}[-]{0,1}[a-z]{0,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}key\\/\", KMSKeyId])"
+    "/properties/MasterUserSecret/KmsKeyId": "$join([\"arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:kms:[a-z]{2}[-]{1}[a-z]{3,10}[-]{0,1}[a-z]{0,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}key\\/\", KmsKeyId])",
```
moataz-mhmd, line 450: What will happen if customer entered KmsKey full Arn instead of KmsKey simple alias? Do CFN neglect applying `propertyTransform` function as if fails to match first regex expression in join?

khebul: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-drift.html

at the moment this is considered a platform limitation

osdrv: @moataz-mhmd from the documentation (https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-model-false-drift.html): 

```
The property transform provides a way for CloudFormation to accurately compare the resource property value specified in the template with the value returned from the read handler. During drift detection, if CloudFormation finds a property where the template value differs from the value returned by the read handler, it determines if a property transform has been defined for that property in the resource schema. If it has, CloudFormation applies that property transform to the value specified in the template, and then compares it to the read handler value again. If these two values match, the property isn't considered to have drifted, and is marked as IN_SYNC
```

I.e. it first would compare the actual value (expanded ARN) and it is expected to match immediately. I.e. propertyTransform should not be engaged at all.

moataz-mhmd: So If I understood it correctly, We consider these two inputs are equal
i.g:
```
KmsKeyId: mySecret1
MasterUserSecret:
   KmsKeyId: arm:aws:us-east-1:kms:antherSecret
```  
is equal to 

```
KmsKeyId: mySecret1
MasterUserSecret:
   KmsKeyId: arm:aws:us-east-1:kms:mySecret1
```

osdrv: @moataz-mhmd not really. The KmsKey alias drift is a known case.

moataz-mhmd (approved):

PR #392, opened 2023-01-31
[DBClusterParameterGroup] Limit max filter size to 100

osdrv:
This commit limits the parameter name filter size to 100 elements per request. RDS API would reject a request containing more than 100 parameter names in the filter. The commit removes the recursive progressChain tailing and proposes a simplistic sequential current parameter accumulator buildup.

### DIFF
--- a/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
+++ b/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
@@ -1,6 +1,7 @@
 package software.amazon.rds.dbclusterparametergroup;
 
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
```diff
@@ -225,36 +237,47 @@ private ProgressEvent<ResourceModel, CallbackContext> validateModelParameters(fi
...
+            try {
+                do {
+                    final DescribeDbClusterParametersResponse response = fetchDbClusterParameters(proxyClient, model, parameterNamePartition, marker);
+                    for (final Parameter parameter : response.parameters()) {
+                        parametersAcc.put(parameter.parameterName(), parameter);
                     }
-                    return ProgressEvent.progress(resourceModel, context);
-                });
+                    marker = response.marker();
+                } while (marker != null);
```
moataz-mhmd, line 262: Optional: we can add MAX_DEPTH parameter to avoid infinite loop in case service returned constant marker as flaw?

moataz-mhmd (approved):

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
osdrv: Thanks, will add it

Outige (approved):

PR #394, opened 2023-02-01
[DBParameterGroup] Limit max parameter name filter size to 100

osdrv:
This commit limits the parameter name filter size to 100 elements per request. RDS API would reject a request containing more than 100 parameter names in the filter. The commit removes the recursive progressChain tailing and uses a simplistic sequential current parameter accumulator buildup.

This commit is a follow-up on https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/392.

### DIFF
--- a/aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/BaseHandlerStd.java
+++ b/aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/BaseHandlerStd.java
@@ -1,23 +1,24 @@
 package software.amazon.rds.dbparametergroup;
 
-import java.util.Collection;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import com.amazonaws.util.StringUtils;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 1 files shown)


moataz-mhmd (approved):

aws-rds-dbparametergroup/src/main/java/software/amazon/rds/dbparametergroup/BaseHandlerStd.java
```diff
@@ -381,63 +385,95 @@ private Map<String, Parameter> getParametersToReset(
...
-                    currentParams.putAll(response.parameters().stream()
-                            .collect(Collectors.toMap(Parameter::parameterName, Function.identity())));
-                    if (StringUtils.isNullOrEmpty(response.marker())) {
-                        return ProgressEvent.progress(model, context);
+        for (final List<String> paramNamePartition : Lists.partition(paramNames, MAX_PARAMETER_FILTER_SIZE)) {
+            String marker = null;
+            try {
+                int page = 1;
+                do {
+                    if (page > MAX_PARAMETER_DESCRIBE_DEPTH) {
```
khebul, line 399: This effectively limits RDS CFN at 2000 parameters, if default values are used. While I don't think we have any engine with this many parameters yet, I think it is possible that we have this many in the future. I.e. postgres13 has ~250 parameters while postgres14 more than doubled that with over 550.

IMO, potential bugs in AWS RDS will explode other tools like AWS CLI and it will trigger fast error correction from AWS RDS team, which makes this defensive code relatively limited in functionality.

khebul (approved):

PR #414, opened 2023-03-06
[DBClusterParameterGroup] ReadHandler Includes Parameters

osdrv:
This commit adds `parameters` attribute to the response of the `ReadHandler`. The current implementation marks this attribute as write-only, which disqualifies the attribute for a drift detection.

As a side effect, this code change includes a deep refactoring in this handler unit test suite. The refactoring promotes a wide use of the standard testing utility `test_handleRequest_base`.


aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
```diff
@@ -238,57 +254,159 @@ private ProgressEvent<ResourceModel, CallbackContext> validateModelParameters(
...
+            final ProxyClient<RdsClient> proxyClient,
+            final DescribeDbClusterParametersRequest request
+    ) {
+        Iterable<Parameter> result = Collections.emptyList();
+
+        String marker = null;
+        int page = 0;
+        do {
+            if (page >= MAX_DESCRIBE_PAGE_DEPTH) {
+                throw new CfnInvalidRequestException("Max DescribeDBParameters page reached.");
```
moataz-mhmd, line 267: Not sure why this `CfnInvalidRequestException` and not `RuntimeException` like fetch `DefaultEngineVersion`. This error not dependent on customer input. Customer can not trigger this error via his stack.

moataz-mhmd (commented):
This resource has memory issue before due to high number of parameters. We used iterable before because we were using SDK paginators. Now, We are not using them anymore. We can avoid building extra iterables(list) and put results directly in accumulators maps. This could be simpler and more memory efficient.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
osdrv: @moataz-mhmd regarding the iterables: well, `Iterable` is just an interface, not an implementation. The reason is is kept in is because it allows me to concatenate iterable objects together and hide the details of the pagination. Paginator operations are actually a good help when dealing with large amount of paginated requests, but there is a significant flaw that forced me to avoid using them here in DBClusterParameterGroup and in DBParameterGroup. See https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/414 for more details. Once this ticket is fixed, I would prefer to remove these custom `do while` loops and return back to the paginated requests. Keeping `Iterable` in allows us to replace this code with minimal efforts.
I'm not sure I understand the memory consumption concerns you shared. As I've mentioned, `Iterable` does not dictate the use of a specific of the underlying data structure in the container implementing this interface. In our case the underlying container is `List<Parameter>`. I have serious doubts that a `HashMap` would be a more efficient storage memory-wise and GC-wise. Could you please provide a more specific data to back this concern?

moataz-mhmd: I am not sure how a bad request from RDS API's perspective relates to `CfnInvalidRequestException`. Customer will get `CfnInvalidRequestException` not `RdsInvalidRequest`

moataz-mhmd (approved):

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
moataz-mhmd: Got it. You want to keep it refactored in way once the bug resolved we rewrite it again with paginators.

khebul: One more note: Runtime exception needs to be mapped to CFN error code. I think throwing very generic exception can prevent us from doing so. `CfnInvalidRequestException` is by default mapped to `InvalidRequest` and terminate, meanwhile` RuntimeException` will be converted into `InternalError` and retried.

Therefore I think `CfnInvalidRequestException` serves our purposed better at this point in time.

aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
```diff
@@ -359,12 +494,29 @@ private ProgressEvent<ResourceModel, CallbackContext> modifyParameters(
...
-                .injectCredentialsAndInvokeIterableV2(Translator.describeDbClustersRequest(), proxyClient.client()::describeDBClustersPaginator)
-                .stream()
-                .flatMap(describeDbClustersResponse -> describeDbClustersResponse.dbClusters().stream())
-                .filter(dbCluster -> dbCluster.dbClusterParameterGroup().equalsIgnoreCase(model.getDBClusterParameterGroupName()))
-                .allMatch(dbCluster -> dbCluster.status().equals(AVAILABLE));
+        String marker = null;
+        int page = 1;
+        final DescribeDbClustersRequest request = Translator.describeDbClustersRequestForDBClusterParameterGroup(model);
+
+        do {
```
khebul, line 501: Code similar to  this one is repeated a number of times between this resource and DBParameterGroup. Does it makes sense to move this part to common?

osdrv: My attempt to implement a generic method for this operation failed due to a limitation of `AwsResponse`. In particular, `marker` is defined as a member of `SDK_FIELDS` in the corresponding response classes and won't be visible at the level of an abstract class like `AwsResponse`. Given we can not enforce an ad-hoc contract, this turns such a method in a set of functional interface object invocations. Given the long-term approach is to get back to the paginators once we have a reasonable solution, I'd say a little copy-paste is better than a non-sense abstraction that is there to hide the existence of a while-loop.

khebul (approved):

Issue #400, opened 2023-02-08
(AWS::RDS::DBInstance) Tracking/Linking of Issue for RDS Drift Detection false-positives


rgoltz:
We are aware that CloudFormation & RDS ServiceTeam is actively working on migration of RDS resource within CFN platform. We facing different CloudFormation Drift Detection false-positives issues, which I'd like to link here as separate comments. By doing so, this issue acts as an overview / master-issue. Once all linked issues are closed, we/AWS should close this issue here as well.

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1511

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1510

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1509

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1503

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1310

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1307

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1003 / https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/390

osdrv:
@rgoltz thanks very much! Really nice summary page.

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1530

rgoltz:
* https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1699

dbayardAWS:
- https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1857

dbayardAWS:
- https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1856

rgoltz:
- https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/2032

keegannnn:
- https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/2275

PR #417, opened 2023-03-13
[OptionGroup] Fix drift on `OptionConfigurations`

osdrv:
This commit addresses a false drift detection on `OptionConfiguration` items. The nature of this drift comes from the RDS API which populates `OptionConfiguration` objects with default `OptionSettings`. I.e. if the stack defines only a few `OptionSettings` within an `OptionConfiguration` object, RDS would populate it with all default values. CFN looses track of these settings and indicates the drift.

The idea behind this change is to hide all settings that are set to defaults. This could cause a counter-drift when the stack defines a set of `OptionSettings` that happened to match the default values. In this case it is an equivalent of not defining these properties at all, hence we consider this inversion to be a reasonable way to address the initial drift.

Refs https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/issues/400.
Fixes https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/1310.


aws-rds-optiongroup/src/main/java/software/amazon/rds/optiongroup/ReadHandler.java
```diff
@@ -45,19 +51,39 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
...
     }
+
+    private List<Option> getOverriddenOptionConfigurations(final OptionGroup optionGroup) {
+        final List<Option> overriddenOptions = new ArrayList<>();
+        if (CollectionUtils.isNotEmpty(optionGroup.options())) {
+            for (final Option option : optionGroup.options()) {
+                final List<OptionSetting> optionSettings = new ArrayList<>();
+                if (CollectionUtils.isNotEmpty(option.optionSettings())) {
+                    for (final OptionSetting optionSetting : option.optionSettings()) {
+                        final String defaultValue = optionSetting.defaultValue();
```
khebul, line 78: what will happen if option is listed in Customer's stack definition, but value is matching the default setting?

moataz-mhmd: If customer got drift in this case, I think that will be a bug. I understand that we can not check stack Options as read handler input will contains only resource identifier without stack attributes. We might need CFN team assistance to provide a good way to handle default attributes.

khebul (approved):

PR #429, opened 2023-04-11
[DBInstance] Set performance insights key on modify-after-create

osdrv:
This commit changes ModifyAfterCreate method of the Translator class. The method is instructed to pass the performance insights(PI) kms key ID if PI is enabled. RDS API expects the key ID upon every modify request if the instance was created with PI enabled and a key ID was provided.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
@@ -531,6 +531,7 @@ public class Translator {
                 .maxAllocatedStorage(model.getMaxAllocatedStorage())
                 .monitoringInterval(model.getMonitoringInterval())
                 .monitoringRoleArn(model.getMonitoringRoleArn())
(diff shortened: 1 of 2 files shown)


moataz-mhmd (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -534,6 +534,10 @@ public static ModifyDbInstanceRequest modifyDbInstanceAfterCreateRequest(final R
                 .preferredBackupWindow(model.getPreferredBackupWindow())
                 .preferredMaintenanceWindow(model.getPreferredMaintenanceWindow());
 
+        if (BooleanUtils.isTrue(model.getEnablePerformanceInsights())) {
+            builder.performanceInsightsKMSKeyId(model.getPerformanceInsightsKMSKeyId());
```
khebul, line 538: Which scenario requires us to set this value conditionally?

moataz-mhmd (approved):

PR #437, opened 2023-04-19
[DBInstance] Set source instance engine on read replica

osdrv:
This commit adds an additional source db instance fetch upon a read replica create. At the moment, storage-related attributes are being set upon a create for non-SQLServer engines only. Due to the safeguard in place, the handler treats all unset engine types as SQLServer and hence sets no storage-related attributes on read replicas.

The idea is to set the replica engine type from the source instance so the handler doesn't have to make a guess whether it is a SQLServer engine.


moataz-mhmd (approved):

osdrv:
Superseded by https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/438

PR #227, opened 2022-06-30
[DBSnapshot] create new resource DBSnapshot 

moataz-mhmd:
*Description of changes:*
This changes contains DBSnapshot resource implementation following most recent AWS RDS Documentation: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_CreateSnapshot.html  

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #439, opened 2023-04-19
[DBSnapshot] Create AWS::RDS::DBSnapshot resource

osdrv:
This commit creates a new resource called `AWS::RDS::DBSnapshot`.

Documentation references:
  * https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_CreateSnapshot.html
  * https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_CopySnapshot.html
  * https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_DeleteSnapshot.html

Supersedes https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/227.


moataz-mhmd (commented):
The only concern, This resource is antipattern for CFN resources. CFN resources used to be actual live resource. Snapshot is different in its meaning as it is point of time image of live resource. It is not clear what is snapshot lifecycle.
Open questions:
1. Is it dependent on source instance updates or totally isolated? If isolated, AFAIK, CFN always triggers update for resource whenever their dependencies updated. How we will exempt this resource from this behavior. 
2. Replacing resource can cause data loss. In DBInstance, Default deletion policy is to take snapshot of source instance. What will be the default deletion policy for dbsnapshot resource?
3. In other services like: Dynamo, They do not provide snapshot/backup resource i.e: AWS::DynamoDB::Backup. But they provide  a [BackupPlan](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-backup-backupplan.html).  Should we do something similar here?


I think CloudFormation team might provide more details here about other services and advice if we should have snapshot resource. We can involve @ammokhov in the thread.

PR #460, opened 2023-08-21
[DBInstance] Detect engine before running Create workflow

khebul:
This PR adds engine detection step before any create operations take place. This allows using create workflow instead of create-modify workflow in some cases when customer does not explicitly specify engine with restore-from-snapshot, create-read-replica and restore-to-point-in-time scenarios.

Create-modify workflow remains main method of provisioning any SQL server DBInstances


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -169,12 +180,42 @@ ApiVersion.DEFAULT, safeAddTags(this::createDbInstance)
                             logger
                     );
                 });
-    }
+    }final
```
dbbh, line 183: typo?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -169,12 +180,42 @@ ApiVersion.DEFAULT, safeAddTags(this::createDbInstance)
...
 
     private HandlerMethod<ResourceModel, CallbackContext> safeAddTags(final HandlerMethod<ResourceModel, CallbackContext> handlerMethod) {
         return (proxy, rdsProxyClient, progress, tagSet) -> progress.then(p -> Tagging.safeCreate(proxy, rdsProxyClient, handlerMethod, progress, tagSet));
     }
 
+    private String fetchEngine(final ProxyClient<RdsClient> client, final ResourceModel model) {
```
dbbh, line 189: I understand why this is needed but it would be great to start seeing our handler logic simplify in the long run by enforcing better API standards on the RDS side.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -56,18 +58,49 @@ public static DescribeDbClustersRequest describeDbClustersRequest(final Resource
                 .build();
     }
 
+    public static DescribeDbClustersRequest describeDbClusterByIdentifierRequest(final String dbClusterIdentifier) {
```
dbbh, line 61: NIT: I believe the `ByIndentifierRequest` portion of the signature is made redundant by the parameter list. This however is mostly of preference nature and we can keep the method as it is :)

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -104,6 +137,12 @@ public static CreateDbInstanceReadReplicaRequest createDbInstanceReadReplicaRequ
                 .tags(Tagging.translateTagsToSdk(tagSet))
                 .useDefaultProcessorFeatures(model.getUseDefaultProcessorFeatures())
                 .vpcSecurityGroupIds(CollectionUtils.isNotEmpty(model.getVPCSecurityGroups()) ? model.getVPCSecurityGroups() : null);
+        if (!ResourceModelHelper.isSqlServer(model)) {
```
dbbh, line 140: We need to be very defensive here and make sure we have the coverage for both code paths -> IsSqlServer and IsNotSqlServer on our testing.

dbbh: Same for the other code paths below with a special callout on `if (ResourceModelHelper.isStorageParametersModified(model) && ResourceModelHelper.isSqlServer(model))` which is seems like a separate user story.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/util/ResourceModelHelper.java
```diff
@@ -31,13 +32,20 @@ public static boolean shouldUpdateAfterCreate(final ResourceModel model) {
...
+                                (isSqlServer(model) && isStorageParametersModified(model)) ||
                                 BooleanUtils.isTrue(model.getManageMasterUserPassword()) ||
                                 BooleanUtils.isTrue(model.getDeletionProtection()) ||
                                 BooleanUtils.isTrue(model.getEnablePerformanceInsights())
                 );
     }
 
+    public static boolean isSqlServer(final ResourceModel model) {
+        final String engine = model.getEngine();
+        // treat unknown engines as SQLServer
```
dbbh, line 44: I believe this benefits from further explanation or guidance as this might not be the case in the future when new engines get released.

aws-rds-dbinstance/src/test/java/software/amazon/rds/dbinstance/TranslatorTest.java
```diff
@@ -316,8 +316,9 @@ public void test_modifyReadReplicaRequest_parameterGroupNotSet() {
     }
 
     @Test
-    public void test_restoreFromSnapshotRequest_storageType_shouldBeSetOnUpdate() {
+    public void test_restoreFromSqlServerSnapshotRequest_storageType_shouldBeSetOnUpdate() {
```
dbbh, line 319: Nice addition on the coverage 👍 
This should help us along and reduce our operational cost over time.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
khebul: Agreed. At this point, however, we need this method in order to make sure all engines use appropriate workflows

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
khebul: Both TranslatorTests (see `test_restoreFromAuroraPostgresSnapshotRequest_storageType_shouldBeSetOnUpdate` and `test_restoreFromSqlServerSnapshotRequest_storageType_shouldBeSetOnUpdate`) and CreateHandlerTests verify SQL server and non SQL server paths.

We also have ResourceModelHelperTest which mostly focus on shouldUpdateAfterCreate method.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/util/ResourceModelHelper.java
khebul: Treating unknown engines is left as a defensive programming. Unless we somehow lost desired model changes, there should not be any situation when we do not know the engine - it is fetched long before this method is accessed. Future engines are unknown at this point in time, I think we should consider them after they are known.

dbbh (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -56,18 +58,49 @@ public static DescribeDbClustersRequest describeDbClustersRequest(final Resource
...
+    public static DescribeDbInstancesRequest describeDbInstanceByDBInstanceIdentifierRequest(final String dbInstanceIdentifier) {
+        return DescribeDbInstancesRequest.builder()
+                .dbInstanceIdentifier(dbInstanceIdentifier)
+                .build();
+    }
+
+    public static DescribeDbInstancesRequest describeDbInstanceByResourceIdRequest(final String dbInstanceIdentifier) {
+        return DescribeDbInstancesRequest.builder()
+                .filters(software.amazon.awssdk.services.rds.model.Filter.builder()
+                        .name("dbi-resource-id").values(dbInstanceIdentifier).build())
```
moataz-mhmd, line 82: I just wanted to confirm if this filter is case insensitive. In case the customer specifies a mixed case for the `DbiResourceId` and we send it as is to the Webservice, will the target DbInstance be returned correctly? This is because the Webservice always converts all letters to lowercase for identifiers.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/util/ResourceModelHelper.java
moataz-mhmd, line 44: From my understanding, the detection of an unknown engine as sqlServer will only occur in the UpdateHandler. The `fetchEngine` method will either fill in the engine or fail hard if it is unknown. In my opinion, I would prefer to fail hard and provide a message instructing to add the Engine parameter to the template. Unfortunately, this change is not backward compatible. So, my suggestion is to log this event so that we can track it and potentially make changes in the future.

To achieve this, we can use the logger to log the following message:
```
        requestLogger.log("Unknown engine converted to SqlServer detected",
                "This indicates that we could not extract the dbinstance engine from the given customer template, and we treated it as SqlServer for backward compatibility.");
```

In order to have better accessibility for the RequestLogger, a change has been introduced in the following pull request: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/pull/459/files. To utilize this change, you will need to pass the RequestLogger as a parameter to the method, as the logger is only given to the ResourceHandler during execution.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -175,6 +186,36 @@ private HandlerMethod<ResourceModel, CallbackContext> safeAddTags(final HandlerM
...
```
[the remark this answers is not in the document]
khebul: To be added later

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
khebul: The filter is not case sensitive, customers are free to use whichever case they prefer.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/util/ResourceModelHelper.java
khebul: This method is not accessed at all in UpdateHandler. Let's consider this enhancement in the logging improvement request.

moataz-mhmd: Could you please provide more details on how the engine will remain unknown even after the `fetchEngine` method is called at the beginning of the CreateHandler?

moataz-mhmd (approved):

dbbh (approved):

PR #461, opened 2023-08-24
[Common][All] Update soft failure on stack level tags to `Unauthorize…

moataz-mhmd:
*Description of changes:*

In order to maintain compatibility with older versions, we used to overlook the AccessDenied exception when a customer lacked the permission to add stack level tags. This PR aims to rectify this behavior by following to the latest CloudFormation guidelines and instead throwing an `UnauthorizedTaggingOperation` error code. 

However, we still have the tagging logic in place to distinguish between access denied for resource tags and access denied for stack level tags. Once CloudFormation confirms that it is safe to convert `UnauthorizedTaggingOperation` to the regular `AccessDenied` error code, we can safely remove all the logic related to tag error management and the `safeCreate` mechanism.

### DIFF
--- a/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
+++ b/aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
@@ -50,6 +50,7 @@ public final class Tagging {
                     ErrorCode.AccessDenied,
                     ErrorCode.AccessDeniedException
             ).build();
+    public static final String RDS_ADD_TAGS_TO_RESOURCE_ACTION = "rds:AddTagsToResource";
 
     public static TagSet exclude(final TagSet from, final TagSet what) {
         final Set<Tag> systemTags = new LinkedHashSet<>(from.getSystemTags());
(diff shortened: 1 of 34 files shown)


aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
```diff
@@ -30,21 +30,21 @@
 import software.amazon.rds.common.error.ErrorStatus;
 
 public final class Tagging {
-    public static final ErrorRuleSet SOFT_FAIL_IN_PROGRESS_TAGGING_ERROR_RULE_SET = ErrorRuleSet
+    public static final ErrorRuleSet IGNORE_LIST_TAGS_PERMISSION_DENIED_ERROR_RULE_SET = ErrorRuleSet
```
khebul, line 33: Is it ListTagsForResource permission missing from the call or is it AddTagsForResource?

moataz-mhmd: This is only used in read handlers for ListTagsForResource. Unfortunately, the RDS APIs for describing are not consistent. The DescribeDBInstance and DescribeDBCluster APIs return tags in the response, but other APIs like DescibeDBClusterEndpoint and DescribeDBParameterGroup do not. Instead, we need to make a separate call to ListTagsForResource to retrieve the tags. If this call fails due to permission denied, we ignore it to prevent the entire read handler operation from failing. As a result, the tags will not be included because the customer does not have permission to see them. We will remove this workaround once the Describe* RDS APIs are consistent and all of them return tags in the Describe* response.

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
```diff
@@ -184,56 +185,52 @@ private static RemoveTagsFromResourceRequest removeTagsFromResourceRequest(
...
         if (tagsToAdd.getResourceTags().isEmpty() && tagsToRemove.getResourceTags().isEmpty()) {
-            return softFailErrorRuleSet;
+            return stackTagsErrorRuleSet;
         }
-        return hardFailErrorRuleSet;
+        return resourceTagsErrorRuleSet;
     }
 
-    public static <M, C extends TaggingContext.Provider> ProgressEvent<M, C> safeCreate(
+    public static <M, C extends TaggingContext.Provider> ProgressEvent<M, C> createWithUnauthorizedTagging(
```
khebul, line 210: nit: the name of this method implies that we are creating the resource while unauthorized. We are performing creation of the resource while also using tagging mechanism fallback. `createWithTaggingFallback`?

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
```diff
@@ -184,56 +185,52 @@ private static RemoveTagsFromResourceRequest removeTagsFromResourceRequest(
...
-        });
+        }
+        allTagsResult.getCallbackContext().getTaggingContext().setAddTagsComplete(true);
+        return allTagsResult;
+    }
+
+    private static <M, C extends TaggingContext.Provider> boolean isUnauthorizedTaggingFailure(final ProgressEvent<M, C> allTagsResult,
+                                                                                               final TagSet allTags) {
+        return HandlerErrorCode.AccessDenied.equals(allTagsResult.getErrorCode()) &&
+                        allTags.getResourceTags().isEmpty() &&
```
khebul, line 231: Does the model differ only by resource tags on second execution?

aws-rds-dbcluster/src/test/java/software/amazon/rds/dbcluster/CreateHandlerTest.java
```diff
@@ -169,55 +169,39 @@ public void handleRequest_CreateDbCluster_ServerlessV2ScalingConfiguration() {
...
                 .thenThrow(
                         RdsException.builder()
+                                .message("Role not authorized to execute rds:AddTagsToResource")
                                 .awsErrorDetails(AwsErrorDetails.builder()
                                         .errorCode(ErrorCode.AccessDeniedException.toString())
                                         .build()
-                                ).build())
-                .thenReturn(CreateDbClusterResponse.builder().build());
-        when(rdsProxy.client().addRoleToDBCluster(any(AddRoleToDbClusterRequest.class)))
-                .thenReturn(AddRoleToDbClusterResponse.builder().build());
```
khebul, line 182: Do we need to add a test for AccessDenied failure during CreateDBCluster/AddRoleToDBCluster operations?

Original test verified that AccessDenied during soft-fail tagging did not break the creation of the resource unintentionally. We might want to check that Acces denied during CreateDBCluster/AddRoleToDBCluster operations actually has the intended effect.

I think this is the case as well for RestoreToPointInTime and RestoreFromSnapshot scenarios

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
moataz-mhmd: There won't be another execution following this PR. However, if you're referring to whether a second execution occurred when the resource tags were empty and an access denied error occurred, then the answer is yes.

aws-rds-dbcluster/src/test/java/software/amazon/rds/dbcluster/CreateHandlerTest.java
moataz-mhmd: Other tests are covering these scenario such as `AddRoleToDBCluster` scenario

aws-rds-cfn-common/src/main/java/software/amazon/rds/common/handler/Tagging.java
moataz-mhmd: Done

dbbh (approved):

PR #464, opened 2023-09-05
[DBClusterParameterGroup] Conditionally reset parameters

khebul:
This PR changes ResetParameter logic to only reset changed parameters in DBClusterParameterGroup. This allows us to avoid unnecessary reboots caused by resetting parameters with `pending-reboot` apply method even when unchanged. It also allows to avoid update failures when attempting to reset parameters such as `lower_case_table_names` - which are not safe to reset.

### DIFF
--- a/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
+++ b/aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
@@ -2,6 +2,7 @@ package software.amazon.rds.dbclusterparametergroup;
 
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


aws-rds-dbclusterparametergroup/src/main/java/software/amazon/rds/dbclusterparametergroup/BaseHandlerStd.java
```diff
@@ -451,13 +447,24 @@ static Map<String, Parameter> computeModifiedDBParameters(
...
             final ProgressEvent<ResourceModel, CallbackContext> progress,
             final AmazonWebServicesClientProxy proxy,
-            final ProxyClient<RdsClient> proxyClient
+            final ProxyClient<RdsClient> proxyClient,
+            final Map<String, Parameter> currentParameters,
+            final Map<String, Object> desiredParams
     ) {
+        // reset only parameters that are missing from desired params. If parameter is in desired param, its value will be updated anyway.
+        final Map<String, Parameter> toBeReset = currentParameters.keySet().stream()
+                .filter(p -> !desiredParams.containsKey(p))
```
moataz-mhmd, line 459: There is no need to reset all parameters except for the desired ones. We can only reset parameters that are not included in the desired ones and have been changed by the user. which `Source` will be `user`. Resetting a parameter that has the `Source` as 'Engine default' is pointless.
Docs: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_Parameter.html

khebul: The code above has the same effect. We only reset parameters which are missing from desired ones, but were present in current parameters retrieved via filtered describe call.

moataz-mhmd: So, if I understood correctly, after this change, we will no longer undo any changes made by the customer to parameters through the UI console or any other client (out of band). This is because these parameters are not found in the previous or desired state.

khebul: Yes, just like in DBParameterGroup

moataz-mhmd (approved):

dbbh (approved):

PR #465, opened 2023-09-19
Add iam CreateServiceLinkedRole permissions

dbbh:
*Issue #, if available:*

*Description of changes:*
Cloudformation handlers need to specify which permissions are required to perform the operations requested by customers. This CR adds CreateServiceLinkedRole on create for other resources similar to what is being done for DbInstance

### DIFF
--- a/aws-rds-dbcluster/aws-rds-dbcluster.json
+++ b/aws-rds-dbcluster/aws-rds-dbcluster.json
@@ -462,6 +462,7 @@
   "handlers": {
     "create": {
       "permissions": [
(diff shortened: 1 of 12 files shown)


khebul (approved):

moataz-mhmd (changes requested):
Unfortunately, this change will not work. The permissions specified in the schema are crucial for contract tests. However, Cloudformation relies on the `resource-role.yml` file, which combines all the permissions from the resource schema. In order to update this file, you will have to execute the `cfn generate` command for each resource.

moataz-mhmd (approved):

PR #469, opened 2023-10-05
[DBInstance] Use DBParameterGroup in CreateReadReplica for MySql engine

moataz-mhmd:
*Description of changes:*
This code update addresses the following issues:
- Sets the DBParameterGroup in the CreateReadReplica request for the MySql Engine.
- It handles the scenario to fill engine where the customer does not provide the Engine parameter and the source instance/cluster is in a different region.
- It fixes a bug where attempting to fetch the sourceDBInstance or sourceDBCluster in the same region using its ARN was causing an error. This was because calling the describe function with an identifier longer than 63 characters always fails.
- It now uses the LoggingProxyClient to wrap the creation of a new client, allowing it to emit executed calls in our CloudWatch logs.


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -210,10 +216,29 @@ private String fetchEngine(final ProxyClient<RdsClient> client, final ResourceMo
         }
 
         if (ResourceModelHelper.isDBInstanceReadReplica(model)) {
-            return fetchDBInstance(client, model.getSourceDBInstanceIdentifier()).engine();
+            final String sourceDBInstanceArn = model.getSourceDBInstanceIdentifier();
+            final String sourceDBInstance = ResourceModelHelper.isValidArn(sourceDBInstanceArn) ?
```
dbbh, line 220: I think we could be a bit more defensive here. If a customer passes on a malformed ARN we are treating it just like the customer is passing an actual DBInstance ID. The concern is that in the future we could simply look at the variable name and make wrong assumptions. Perhaps naming it as sourceDBInstanceIdOrArn would make it explicit for future developers to avoid assumption traps.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -210,10 +216,29 @@ private String fetchEngine(final ProxyClient<RdsClient> client, final ResourceMo
...
+                    ResourceModelHelper.getResourceNameFromArn(sourceDBInstanceArn) : sourceDBInstanceArn;
+            if (ResourceModelHelper.isCrossRegionDBInstanceReadReplica(model, currentRegion)) {
+                final String sourceRegion = ResourceModelHelper.getRegionFromArn(sourceDBInstanceArn);
```
dbbh, line 223: Just a few lines above we are checking if the ARN is valid. What happens if a malformed ARN is sent to the method?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -210,10 +216,29 @@ private String fetchEngine(final ProxyClient<RdsClient> client, final ResourceMo
...
+                final ProxyClient<RdsClient> sourceRegionClient = new LoggingProxyClient<>(logger,
```
dbbh, line 224: Are we instantiating the client multiple times? For both cases we are using the source region for the client.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -152,6 +152,11 @@ public static CreateDbInstanceReadReplicaRequest createDbInstanceReadReplicaRequ
             builder.storageThroughput(model.getStorageThroughput());
             builder.storageType(model.getStorageType());
         }
+
+        if (ResourceModelHelper.isMySQL(model)) {
```
dbbh, line 156: This line stands out. I understand this is because of particularities of the engine but we should document this somewhere so that we can revisit this code in the future. Perhaps in the PR description we could add some context of why this is needed.

dbbh (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
moataz-mhmd: If It is cross-region read replica dbinstance, the code will return the fetched engine without instantiating the client again. Therefore, it will never be instantiated again in DBCluster case.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
moataz-mhmd: `isCrossRegionDBInstanceReadReplica` function is to validate the correctness of an ARN and retrieve its region. This region then compared with the current region. If the comparison yields false, the logic will not continue. Therefore, if the ARN is malformed, this logic will not be executed.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
moataz-mhmd: Done.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -152,6 +152,11 @@ public static CreateDbInstanceReadReplicaRequest createDbInstanceReadReplicaRequ
...
+            builder.dbParameterGroupName(model.getDBParameterGroupName());
```
khebul, line 157: If dbparameter group is set unconditionally, what effect would that have on non-mysql engines?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/util/ResourceModelHelper.java
```diff
@@ -48,6 +48,10 @@ public static boolean isSqlServer(final ResourceModel model) {
         return engine == null || engine.contains(SQLSERVER_ENGINE);
     }
 
+    public static boolean isMySQL(final ResourceModel model) {
+        final String engine = model.getEngine();
+        return engine != null && engine.startsWith(MYSQL_ENGINE_PREFIX);
```
khebul, line 53: Are `aurora-mysql` and `aurora` engines considered MYSQL engine? If yes, we need to conisder that too

dbbh (dismissed):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/util/ResourceModelHelper.java
khebul: The link you have posted also mentions Oracle DB instances. Shall we use this parameter in that case too?

dbbh (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/CreateHandler.java
```diff
@@ -70,6 +72,7 @@ protected ProgressEvent<ResourceModel, CallbackContext> handleRequest(
         final ResourceModel model = request.getDesiredResourceState();
         final Collection<DBInstanceRole> desiredRoles = model.getAssociatedRoles();
         final boolean isMultiAZ = BooleanUtils.isTrue(model.getMultiAZ());
+        callbackContext.setCurrentRegion(request.getRegion());
```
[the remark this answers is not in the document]
moataz-mhmd: Yes, CFN itself provides it. Without this provision, the condition would not be met and there would be no change in the current behavior prior to merging the PR. Then we will contact CFN to inquire why they did not provide it in this case.

dbbh (approved):

khebul (approved):

moataz-mhmd:
The change has been merged and deployed globally. If you're still experiencing any issues, please don't hesitate to reach out.

PR #479, opened 2023-10-31
[DBInstance] Ignore terminal domain on delete

dbbh:
*Issue #, if available:*

*Description of changes:*
This change moves the terminal state check in the DBInstance handler for RDS resources into a separate helper method. Creates and modifies continue to assert terminal status on all resources while delete requests are now only checking for RDS resources. This means terminal domains will no longer impact delete workflows.


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -656,9 +656,13 @@ private void assertNoDomainMembershipTerminalStatus(final DBInstance dbInstance)
     }
 
     private void assertNoTerminalStatus(final DBInstance dbInstance) throws CfnNotStabilizedException {
+        assertNoTerminalStatusOnRdsResources(dbInstance);
+        assertNoDomainMembershipTerminalStatus(dbInstance);
+    }
+
+    private void assertNoTerminalStatusOnRdsResources(final DBInstance dbInstance) throws CfnNotStabilizedException {
```
moataz-mhmd, line 663: Would it be advisable to include a unit test for this scenario to avoid the possibility of accidentally reintroducing the removed condition? This way, developers will be aware of its significance of it in the delete scenario.

dbbh: Yes we need unit tests, not only for this feature but for many other helper methods inside the base handler. The test class need a small refactoring to take in both the proxy client and the handler to be able to test for these methods, this one included.

moataz-mhmd (approved):

PR #481, opened 2023-11-09
[CustomDBEngineVersion] Update ListHandler logic to return only created custom

moataz-mhmd:
*Description of changes:*
In this code modification, I have adjusted the behavior to only return the custom engine versions that have been created, rather than returning all engine versions present in RDS.

### DIFF
--- a/aws-rds-customdbengineversion/src/main/java/software/amazon/rds/customdbengineversion/ListHandler.java
+++ b/aws-rds-customdbengineversion/src/main/java/software/amazon/rds/customdbengineversion/ListHandler.java
@@ -1,7 +1,11 @@
 package software.amazon.rds.customdbengineversion;
 
+import java.util.Collection;
 import java.util.List;
 
+import org.apache.commons.lang3.StringUtils;
+
+import com.google.common.collect.ImmutableList;
 import software.amazon.awssdk.services.rds.RdsClient;
 import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
 import software.amazon.cloudformation.proxy.Logger;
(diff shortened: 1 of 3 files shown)


aws-rds-customdbengineversion/src/main/java/software/amazon/rds/customdbengineversion/ListHandler.java
```diff
@@ -13,6 +17,14 @@
 
 public class ListHandler extends BaseHandlerStd {
 
+    final static Collection<String> CUSTOM_ENGINES = ImmutableList.of(
```
khebul, line 20: Have you considered client side filter for custom engine versions? CFN in that case will not need to be updated when new custom engine is added.

moataz-mhmd: - If we implement client-side filtering, it means that we will be sending empty intermediate pages back to the ListHandler caller. Please note that the ListHandler only returns one page per invocation, not all of the collected data. 
- DescribeEngineVersions call is quite intensive in terms of response size and latency. It would be better to implement server-side filtering to avoid this.

aws-rds-customdbengineversion/src/main/java/software/amazon/rds/customdbengineversion/ListHandler.java
```diff
@@ -13,6 +17,14 @@
...
+            "custom-oracle-ee",
+            "custom-oracle-ee-cdb",
+            "custom-sqlserver-ee",
+            "custom-sqlserver-se",
+            "custom-sqlserver-web"
```
dbbh, line 25: This list is likely to change over time. We need to have an action item to get these values dynamically.

moataz-mhmd: True, We are going to ask for a new API for this purpose from RDS service.

aws-rds-customdbengineversion/src/main/java/software/amazon/rds/customdbengineversion/Translator.java
```diff
@@ -110,13 +114,17 @@ static ResourceModel translateFromSdk(
...
                                                                                   final String nextToken) {
         return DescribeDbEngineVersionsRequest.builder()
                 .engine(model.getEngine())
                 .engineVersion(model.getEngineVersion())
-                .marker(nextToken)
+                .filters(Filter.builder()
+                        .name(ENGINE_FILTER_NAME)
+                        .values(engines)
+                        .build())
                 .includeAll(true)
```
dbbh, line 126: `Specifies whether to also list the engine versions that aren't available. The default is to list only available engine versions.`

dbbh (approved):

PR #484, opened 2023-11-15
[DBInstance/DBCluster] Add CreateDB[Cluster]Snapshot permissions on d…

dbbh:
…elete handlers

*Issue #, if available:*

*Description of changes:*
This change includes rds:CreateDBSnapshot permissions on the DBInstance delete handler and rds:CreateDBClusterSnapshot on DBCluster delete handler for final snapshot creation on deletes.


moataz-mhmd (approved):

moataz-mhmd (changes requested):
Unfortunately, this will not be effective. CloudFormation specifically looks for the necessary permissions in the `resource-role.yaml` file. Therefore, if you make any changes to the schema permissions, you will need to run the `cfn generate` command in order to update the corresponding `resource-role.yaml` file for the Uluru resource.

moataz-mhmd (approved):

PR #490, opened 2023-12-15
[CustomDBEngineVersion]Rac rolling os patching

dbbh:
*Issue #, if available:*

*Description of changes:*
This PR adds support for RAC rolling OS patching.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,13 @@
 aws-rds-handlers.iml
 rpdk.log
+/build/
+# python helper scripts
+bin/
+lib/
+env/
 # for IntelliJ IDEA
 .idea/
+# Gradle
+.gradle
+/gradle/
+/wrapper/
--- a/aws-rds-cfn-common/pom.xml
+++ b/aws-rds-cfn-common/pom.xml
@@ -28,12 +28,12 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>utils</artifactId>
-            <version>2.21.17</version>
+            <version>2.24.13</version>
         </dependency>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>rds</artifactId>
-            <version>2.21.17</version>
+            <version>2.24.13</version>
         </dependency>
         <dependency>
             <groupId>software.amazon.cloudformation</groupId>
@@ -190,7 +190,7 @@
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <version>0.8.4</version>
+                <version>0.8.8</version>
                 <configuration>
                     <excludes>
                     </excludes>
--- a/aws-rds-cfn-common/src/test/java/software/amazon/rds/common/handler/TaggingTest.java
+++ b/aws-rds-cfn-common/src/test/java/software/amazon/rds/common/handler/TaggingTest.java
@@ -40,6 +40,7 @@ import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
 import software.amazon.rds.common.error.ErrorRuleSet;
 import software.amazon.rds.common.error.ErrorStatus;
 import software.amazon.rds.common.error.UnexpectedErrorStatus;
+import software.amazon.rds.common.handler.Tagging.TagSet;
 import software.amazon.rds.common.logging.LoggingProxyClient;
 import software.amazon.rds.common.logging.RequestLogger;
 import software.amazon.rds.common.printer.FilteredJsonPrinter;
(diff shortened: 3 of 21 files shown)


aws-rds-customdbengineversion/aws-rds-customdbengineversion.json
```diff
@@ -122,14 +134,23 @@
...
+    "/properties/UseAwsProvidedLatestImage"
   ],
   "handlers": {
     "create": {
       "permissions": [
+        "ec2:CopySnapshot",
+        "ec2:DeleteSnapshot",
+        "ec2:DescribeSnapshots",
         "kms:CreateGrant",
+        "kms:Decrypt,",
```
0ooopsFibby, line 150: Type of having ',' inside

aws-rds-customdbengineversion/resource-role.yaml
```diff
@@ -30,8 +30,14 @@ Resources:
             Statement:
               - Effect: Allow
                 Action:
+                - "ec2:CopySnapshot"
+                - "ec2:DeleteSnapshot"
+                - "ec2:DescribeSnapshots"
                 - "kms:CreateGrant"
+                - "kms:Decrypt,"
```
0ooopsFibby, line 37: Additional ',' inside Decrypt

aws-rds-customdbengineversion/aws-rds-customdbengineversion.json
```diff
@@ -103,12 +115,12 @@
...
   },
   "required": [
-    "DatabaseInstallationFilesS3BucketName",
     "Engine",
     "EngineVersion"
   ],
   "writeOnlyProperties": [
-    "/properties/Manifest"
+    "/properties/Manifest",
+    "/properties/UseAwsProvidedLatestImage"
```
0ooopsFibby, line 123: /SourceCustomDbEngineVersionIdentifier is also not allowed for modification.

aws-rds-customdbengineversion/aws-rds-customdbengineversion.json
dbbh: Thanks for catching this typo. Updating the PR

aws-rds-customdbengineversion/aws-rds-customdbengineversion.json
dbbh: Yes, those properties were added as `createOnlyProperties`. `writeOnlyProperties` are not returned as part of the describe and thus are exempt from drift detection. With `createOnlyProperties` it means changing this value will trigger a resource replacement as we can't modify it so we create a new resource with the new value.

aws-rds-customdbengineversion/aws-rds-customdbengineversion.json
```diff
@@ -76,6 +76,18 @@
       "type": "string",
       "description": "The ARN of the custom engine version."
     },
+    "SourceCustomDBEngineVersionIdentifier": {
+      "type": "string",
+      "description": "The identifier of the source custom engine version."
+    },
+    "UseAwsProvidedLatestImage": {
```
moataz-mhmd, line 83: This parameter may cause issues as it can be easily applied during the creation process, but problems may arise when modifying CEV. Customers will expect that updating their CFN stack with the CEV resource will trigger an update to keep their CEV up to date with the latest image at the time of the update. However, the truth is that this parameter will be ignored and the CEV will not be updated. On the other hand, if the customer makes any changes that require replacement and are not related to the image ID, such as changing the name of the CEV, the CEV image will be updated, resulting in inconsistency.

To resolve this issue, the CEV describe or RDS WS should provide an easy way to determine the latest image version. This will allow the UpdateHandler to determine if an upgrade is needed or not during invocation.

aws-rds-customdbengineversion/aws-rds-customdbengineversion.json
```diff
@@ -103,12 +115,13 @@
...
   "required": [
-    "DatabaseInstallationFilesS3BucketName",
     "Engine",
     "EngineVersion"
   ],
   "writeOnlyProperties": [
-    "/properties/Manifest"
+    "/properties/Manifest",
+    "/properties/SourceCustomDbEngineVersionIdentifier",
+    "/properties/UseAwsProvidedLatestImage"
```
moataz-mhmd, line 124: The parameter `UseAwsProvidedLatestImage` is conditional create only. It should not be taken into account if the current CEV Image matches the latest image. This parameter comes into play when there is a newer image available. In order to avoid confusion, we should consider making some design decisions regarding this parameter. One option could be renaming it to `UseAwsProvidedLatestImageAtCreate` to clarify its purpose or making it intelligent enough to automatically upgrade the engine when a newer image is available. Alternatively, we could explore the possibility of specifying two separate flags, one for creation and one for update. Ultimately, the solution we choose should make the most sense from the customer's perspective.

aws-rds-customdbengineversion/src/main/java/software/amazon/rds/customdbengineversion/Translator.java
```diff
@@ -31,9 +31,12 @@ public static CreateCustomDbEngineVersionRequest createCustomDbEngineVersionRequ
                 .description(model.getDescription())
                 .engine(model.getEngine())
                 .engineVersion(model.getEngineVersion())
+                .imageId(model.getImageId())
                 .kmsKeyId(model.getKMSKeyId())
                 .manifest(model.getManifest())
+                .sourceCustomDbEngineVersionIdentifier(model.getSourceCustomDBEngineVersionIdentifier())
```
moataz-mhmd, line 37: Could you please verify if there have been any recent additions to the CEV statuses for restoration, such as the "restoring" status? If there is a new status, it will need to be added to the followingenum: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-customdbengineversion/src/main/java/software/amazon/rds/customdbengineversion/CustomDBEngineVersionStatus.java . Additionally, please ensure that the new status is marked as "Transient" to ensure successful stabilization logic.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -588,6 +588,7 @@ public static ModifyDbInstanceRequest modifyDbInstanceAfterCreateRequestV12(fina
 
     public static ModifyDbInstanceRequest modifyDbInstanceAfterCreateRequest(final ResourceModel model) {
         final ModifyDbInstanceRequest.Builder builder = ModifyDbInstanceRequest.builder()
+                .allowMajorVersionUpgrade(model.getAllowMajorVersionUpgrade())
```
moataz-mhmd, line 591: I am unsure if this change is intended to be included in this pull request, as it is not mentioned in the pull request description. Could you please confirm and update the pull request description? Alternatively, it may be better to create a separate pull request for this change. Combining two features in different resources within one pull request could lead to confusion and make it harder to track in the git commit history.

moataz-mhmd (approved):

PR #498, opened 2024-01-16
[DBInstance] Revert isEngineMutable check to use previous model for same engine check

elnaggar-amr:
*Issue #, if available:*

*Description of changes:*
Revert isEngineMutable check to use previous model for same engine check

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
@@ -229,6 +229,11 @@ public class Translator {
             builder.storageThroughput(model.getStorageThroughput());
             builder.storageType(model.getStorageType());
         }
+
+        if (ResourceModelHelper.isOracleCDBEngine(model.getEngine())) {
+            builder.engine(null);
+        }
+
         return builder.build();
     }
 
@@ -388,6 +393,11 @@ public class Translator {
             builder.storageThroughput(model.getStorageThroughput());
             builder.storageType(model.getStorageType());
         }
+
+        if (ResourceModelHelper.isOracleCDBEngine(model.getEngine())) {
+            builder.engine(null);
+        }
+
         return builder.build();
     }
 
@@ -599,7 +609,6 @@ public class Translator {
                 .engineVersion(model.getEngineVersion())
                 .manageMasterUserPassword(model.getManageMasterUserPassword())
                 .masterUserPassword(model.getMasterUserPassword())
-                .masterUserPassword(model.getMasterUserPassword())
                 .masterUserSecretKmsKeyId(model.getMasterUserSecret() != null ? model.getMasterUserSecret().getKmsKeyId() : null)
                 .maxAllocatedStorage(model.getMaxAllocatedStorage())
                 .monitoringInterval(model.getMonitoringInterval())
@@ -614,6 +623,11 @@ public class Translator {
             builder.storageThroughput(model.getStorageThroughput());
             builder.storageType(model.getStorageType());
         }
+
+        if (ResourceModelHelper.isOracleCDBEngine(model.getEngine())) {
+            builder.engine(model.getEngine());
+        }
+
         return builder.build();
     }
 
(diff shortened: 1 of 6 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -599,7 +609,6 @@ public static ModifyDbInstanceRequest modifyDbInstanceAfterCreateRequest(final R
                 .engineVersion(model.getEngineVersion())
                 .manageMasterUserPassword(model.getManageMasterUserPassword())
                 .masterUserPassword(model.getMasterUserPassword())
-                .masterUserPassword(model.getMasterUserPassword())
```
moataz-mhmd, line 602: Can you elaborate why `masterUserPassword` is being removed here?

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -614,6 +623,11 @@ public static ModifyDbInstanceRequest modifyDbInstanceAfterCreateRequest(final R
             builder.storageThroughput(model.getStorageThroughput());
             builder.storageType(model.getStorageType());
         }
+
+        if (ResourceModelHelper.isOracleCDBEngine(model.getEngine())) {
+            builder.engine(model.getEngine());
```
moataz-mhmd, line 628: Is this parameter idempotent? What I mean is, if the instance being restored already had the CDB engine, and the Modify call includes the same CDB engine, will it succeed on the RDS side without making any changes or will it fail with a validation error? We have seen cases of non-idempotent behavior in the past, so we need to take that into consideration.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
elnaggar-amr: it's duplicated on line 611. not sure it was duplicate.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
elnaggar-amr: it's idempotent, I tried that manually. If there are no other changes though, there will be no modify after create as CDB to CDB is not conversion and will be skipped from CFN side.

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
moataz-mhmd: Apology missed it :)

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -599,7 +609,6 @@ public static ModifyDbInstanceRequest modifyDbInstanceAfterCreateRequest(final R
                 .engineVersion(model.getEngineVersion())
```
moataz-mhmd, line 609: From: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/oracle-cdb-converting.html
> When you upgrade your database engine version, you can't change the database architecture in the same operation. Therefore, to upgrade an Oracle Database 19c non-CDB to an Oracle Database 21c CDB, you first need to convert the non-CDB to a CDB in one step, and then upgrade the 19c CDB to a 21c CDB in a separate step.

I think if we try to do the CDB conversion and engine version upgrade from 19c to 21c at the same time, it's gonna be problematic on CFN handler. Can you please confirm if we're good to go with this scenario for both restoring and updating?

moataz-mhmd (commented):
Make sure you take care of the rollback situation when making one-way changes, like CDB conversion. Let me give you an example to explain how we handle rollback during modifications. You can check out AllocatedStorage parameter case here: [GitHub Link](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java#L486-L489)

moataz-mhmd (approved):

dbbh (approved):

PR #513, opened 2024-02-29
[DBCluster] Support add support for StorageThroughput

dbbh:
This code change introduces support for GP3 storage type for Aurora engines. With the the storage types supported on Aurora engines, the database will now return a new StorageThroughput as a read-only parameter which will be set automatically by the backend. For more information, check the public documentation related to this feature:

https://aws.amazon.com/about-aws/whats-new/2022/11/amazon-rds-general-purpose-gp3-storage-volumes/

*Issue #, if available:*

*Description of changes:*
- Adds StorageThroughput as a read-only parameter.
- Adds and refactors unit tests for DBCluster related to StorageThroughput and Iops
- Includes missing stabilization on StorageThroughput for DBInstance resource which is already rolled-out.
- Bumps RDS SDK version to 2.24.13

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,10 @@
 aws-rds-handlers.iml
 rpdk.log
+.github
+/build/
 # for IntelliJ IDEA
 .idea/
+# Gradle
+.gradle
+/gradle/
+/wrapper/
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,10 @@
 aws-rds-handlers.iml
 rpdk.log
-.github
 /build/
+# python helper scripts
+bin/
+lib/
+env/
 # for IntelliJ IDEA
 .idea/
 # Gradle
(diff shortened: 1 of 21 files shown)


moataz-mhmd (changes requested):
In the `translateDbClusterFromSdk` method, I would expect to see this `StorageThroughput` parameter being set. This method is responsible for returning the details of a DB Cluster to CFN, including read-only parameters like `DBClusterArn`. It's important to set these parameters in the method because if they are not set, the customer may encounter a null exception when trying to reference them for any reason, such build condition to create another resource or configuring dependent resource.

If you want to take a look at the code implementation, you can find it here: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/Translator.java#L498-L580

documentation: https://docs.aws.amazon.com/cli/latest/reference/rds/describe-db-clusters.html

moataz-mhmd (dismissed):

moataz-mhmd (approved):

PR #514, opened 2024-03-01
[Integration] Update timeout values to work better with the service.

wbkang:
Integration can take 4+ hours to create if the source database is huge. Likewise, the delete should not take more than 30 minutes, so the timeouts are being adjusted.

*Issue #, if available:* N/A

*Description of changes:*

AWS::RDS::Integration can actually take 4+ hours to create/modify if the source database is huge. Likewise, the deletion should not take more than 30 minutes (as opposed to the default 90 minutes). We are adjusting the timeout values to reflect that.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. - YES

### DIFF
--- a/aws-rds-integration/src/main/java/software/amazon/rds/integration/BaseHandlerStd.java
+++ b/aws-rds-integration/src/main/java/software/amazon/rds/integration/BaseHandlerStd.java
@@ -15,6 +15,7 @@ import software.amazon.cloudformation.proxy.Logger;
 import software.amazon.cloudformation.proxy.ProgressEvent;
 import software.amazon.cloudformation.proxy.ProxyClient;
 import software.amazon.cloudformation.proxy.ResourceHandlerRequest;
+import software.amazon.cloudformation.proxy.delay.Constant;
 import software.amazon.rds.common.error.ErrorRuleSet;
 import software.amazon.rds.common.error.ErrorStatus;
 import software.amazon.rds.common.handler.Commons;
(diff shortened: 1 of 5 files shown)


dbbh (approved):

aws-rds-integration/src/main/java/software/amazon/rds/integration/BaseHandlerStd.java
```diff
@@ -40,6 +42,18 @@ public abstract class BaseHandlerStd extends BaseHandler<CallbackContext> {
     protected static final int MAX_LENGTH_INTEGRATION = 63;
     protected static final int CALLBACK_DELAY = 6;
 
+    /** Huge database resync could potentially take a very long time. */
+    public static final Constant CREATE_UPDATE_DELAY = Constant.of()
+            .delay(Duration.ofSeconds(30))
+            .timeout(Duration.ofHours(8))
```
moataz-mhmd, line 48: I would suggest updating the schema with the timeouts. Here an example: [DBInstance Timeout](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbinstance/aws-rds-dbinstance.json#L592).

wbkang: Oops, I missed this message, I will follow up in a second PR.

PR #528, opened 2024-04-02
[DBInstance] Add support for IO2 storage

khebul:
This request adds support for IO2 storage for DBInstance

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
@@ -549,7 +549,7 @@ public class Translator {
     }
 
     private static Boolean isProvisionedIoStorage(final ResourceModel model) {
-        return isIo1Storage(model) || isGp3Storage(model);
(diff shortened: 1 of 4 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java
```diff
@@ -561,6 +561,11 @@ private static Boolean isIo1Storage(final ResourceModel model) {
                 (StringUtils.isEmpty(model.getStorageType()) && model.getIops() != null);
     }
 
+    private static Boolean isIo2Storage(final ResourceModel model) {
+        return StorageType.IO2.toString().equalsIgnoreCase(model.getStorageType()) ||
+                (StringUtils.isEmpty(model.getStorageType()) && model.getIops() != null);
```
moataz-mhmd, line 566: Can you explain what this condition `(StringUtils.isEmpty(model.getStorageType()) && model.getIops() != null)` means? I noticed that it is used in different types like `isGp3Storage`.

khebul: io1 remnant code. It still needs to be a part of io1 check, but not io2

PR #542, opened 2024-05-30
[DBInstance][DBCluster] Add more terminal states

zrfr:
This change is to add more terminal states so that instance/cluster operations fail fast if these states are reached, instead of waiting for the stabilization process to time out.

Status references:
- https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/accessing-monitoring.html
- https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/accessing-monitoring.html

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/status/DBInstanceStatus.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/status/DBInstanceStatus.java
@@ -9,10 +9,15 @@ public enum DBInstanceStatus implements TerminableStatus {
     Deleting("deleting"),
     Failed("failed", true),
     InaccessibleEncryptionCredentials("inaccessible-encryption-credentials", true),
+    InaccessibleEncryptionCredentialsRecoverable("inaccessible-encryption-credentials-recoverable", true),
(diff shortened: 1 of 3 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/status/DBInstanceStatus.java
```diff
@@ -9,10 +9,15 @@ public enum DBInstanceStatus implements TerminableStatus {
     Deleting("deleting"),
     Failed("failed", true),
     InaccessibleEncryptionCredentials("inaccessible-encryption-credentials", true),
+    InaccessibleEncryptionCredentialsRecoverable("inaccessible-encryption-credentials-recoverable", true),
```
dbbh, line 12:
```
The KMS key used to encrypt or decrypt the DB instance can't be accessed. However, if the KMS key is active, restarting the DB instance can recover it.
```
Is a `reboot-db-instance` service call required to move the instance out of this state? We need to consider if we are inclined to account for this WF or if we prefer to delegate this resolution elsewhere.

dbbh (approved):

PR #556, opened 2024-07-08
[DBClusterParameterGroup] Enforce that "aurora_enhanced_binlog" and "binlog_backup" get bundled in the same ModifyDBParameterGroup request

JavierAbella:
There are combinations of ParameterGroups that ModifyDBParameterGroup expected them come in the same request, as they are related and configure a particular functionality. At the same time, the ModifyDBParameterGroup API expects a maximum of 20 parameters that can be modified in a single request. (https://docs.aws.amazon.com/cli/latest/reference/rds/modify-db-parameter-group.html)

In the case that the parameters included in the CFN template exceed the request limit, we could end up in a condition in which 2 related parameters existing in the CFN template, end up in 2 different requests. We need to ensure that all related parameters are sent in the same request.

### DIFF
--- a/aws-rds-dbclusterparametergroup/src/test/java/software/amazon/rds/dbclusterparametergroup/CreateHandlerTest.java
+++ b/aws-rds-dbclusterparametergroup/src/test/java/software/amazon/rds/dbclusterparametergroup/CreateHandlerTest.java
@@ -8,17 +8,23 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.doAnswer;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.IntStream;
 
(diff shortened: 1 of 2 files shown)


angusy29 (approved):

aws-rds-dbclusterparametergroup/src/test/java/software/amazon/rds/dbclusterparametergroup/CreateHandlerTest.java
```diff
@@ -405,6 +440,77 @@ public void handleRequest_SuccessWithParameters() {
...
+                        .tagList(Tag.builder().key(TAG_KEY).value(TAG_VALUE).build())
+                        .build());
+
+        // Defining 2 parameters that have to be bundled in the same request. We do this by including them in an
+        // insert-ordered Map, and being far away one from each other. With these conditions, they would naturally
+        // be bundled in 2 different API requests to "modify dbclusterparametergroup".
+        ResourceModel resourceModel = RESOURCE_MODEL.toBuilder()
+                .parameters(InsertionOrderedParamBuilder.builder()
+                        .addParameter("aurora_enhanced_binlog", "1")
+                        .addRandomParameters(BaseHandlerStd.MAX_PARAMETERS_PER_REQUEST)
```
ccmlst, line 490: would be good to add hardcoded parameters that should not be bundle and that they existed in 2 separate requests.

PR #557, opened 2024-07-12
fix(DBInstance): handle cases where both previous and desired allocated storage are null

kylenie-aws:
This change addresses an issue causing DB instance updates to fail during rollback with an InternalFailure. The error message was "Cannot invoke 'java.lang.Integer.intValue()' because 'b' is null". The fix ensures the max() function can handle null inputs.

### DIFF
--- a/aws-rds-dbinstance/src/test/java/software/amazon/rds/dbinstance/TranslatorTest.java
+++ b/aws-rds-dbinstance/src/test/java/software/amazon/rds/dbinstance/TranslatorTest.java
@@ -94,6 +94,20 @@ class TranslatorTest extends AbstractHandlerTest {
     }
 
     @Test
+    public void test_modifyDbInstanceRequest_isRollback_NoAllocatedStorage() {
+        final ResourceModel previousModel = RESOURCE_MODEL_BLDR()
+                .allocatedStorage(null)
+                .build();
+        final ResourceModel desiredModel = RESOURCE_MODEL_BLDR()
+                .allocatedStorage(null)
+                .storageType(STORAGE_TYPE_IO1)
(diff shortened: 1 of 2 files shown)


aws-rds-dbinstance/src/test/java/software/amazon/rds/dbinstance/TranslatorTest.java
```diff
@@ -93,6 +93,20 @@ public void test_modifyDbInstanceRequest_isRollback_IncreaseAllocatedStorage() {
...
     }
 
+    @Test
+    public void test_modifyDbInstanceRequest_isRollback_NoAllocatedStorage() {
+        final ResourceModel previousModel = RESOURCE_MODEL_BLDR()
+                .allocatedStorage(null)
+                .build();
+        final ResourceModel desiredModel = RESOURCE_MODEL_BLDR()
+                .allocatedStorage(null)
+                .storageType(STORAGE_TYPE_IO1)
```
ccmlst, line 103: is this required in the test?

ccmlst (approved):

aws-rds-dbinstance/src/test/java/software/amazon/rds/dbinstance/TranslatorTest.java
kylenie-aws: Yes, max() function only involved when it is a ProvisionedIoStorage.
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/Translator.java#L496

PR #559, opened 2024-07-18
fix(DBCluster, DBInstance): remove incorrect defaults from schema

zrfr:
Removing incorrect defaults from the schema.

We noticed that some customers use the default values from the schema to fill in missing values. This was never the intended purpose, but because the defaults were wrong, the incorrect values caused resources to fail to create or update.

### DIFF
--- a/aws-rds-dbcluster/aws-rds-dbcluster.json
+++ b/aws-rds-dbcluster/aws-rds-dbcluster.json
@@ -88,8 +88,7 @@
     },
     "DBClusterParameterGroupName": {
       "description": "The name of the DB cluster parameter group to associate with this DB cluster.",
-      "type": "string",
-      "default": "default.aurora5.6"
+      "type": "string"
     },
     "DBSubnetGroupName": {
       "description": "A DB subnet group that you want to associate with this DB cluster.",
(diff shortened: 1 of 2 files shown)


aws-rds-dbcluster/aws-rds-dbcluster.json
```diff
@@ -175,8 +174,7 @@
     },
     "MonitoringInterval": {
       "description": "The interval, in seconds, between points when Enhanced Monitoring metrics are collected for the DB cluster. To turn off collecting Enhanced Monitoring metrics, specify 0. The default is 0.",
```
angusy29, line 176: The documentation would be wrong if you remove the default.

aws-rds-dbcluster/aws-rds-dbcluster.json
```diff
@@ -175,8 +174,7 @@
...
-      "type": "integer",
-      "default": 0
```
angusy29, line 179: I think we'd need to remove the default for DBInstance as well, since the CLI defaults to 0 for us already: https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-instance.html

ccmlst (approved):

PR #561, opened 2024-07-31
[DBInstance] Add region validation for the attributes AutomaticBackup…

JavierAbella:
AWS::RDS::DBInstance takes an AWS region name as input without validation and passes it directly into the AWS Java SDK v2 client. This can be specified in the AutomaticBackupReplicationRegion, SourceDBInstanceIdentifier, SourceDBClusterIdentifier properties.

The consequence of not validating this field is both a possible security issue where you could try to hit user-specified URLs, and also when it fails, it gives back a non-friendly error response to the client, making it hard to troubleshoot.

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -112,6 +112,7 @@ import software.amazon.rds.dbinstance.status.DomainMembershipStatus;
 import software.amazon.rds.dbinstance.status.OptionGroupStatus;
 import software.amazon.rds.dbinstance.status.ReadReplicaStatus;
 import software.amazon.rds.dbinstance.status.VPCSecurityGroupStatus;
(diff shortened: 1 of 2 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -671,6 +675,19 @@ private void assertNoTerminalStatus(final DBInstance dbInstance) throws CfnNotSt
         assertNoDomainMembershipTerminalStatus(dbInstance);
     }
 
+    /**
+     * There are fields in the request (e.g. SourceDBInstanceIdentifier and SourceDBClusterIdentifier) that accept 2 types
+     * of identifiers: InstanceId or ARN. In the case of containing an ARN, we need to validate that the region is correct,
+     * as it will be used to contract the client endpoint, and a malformed region would lead to both security issues
```
ccmlst, line 681: typo: "contract" instead of "contact".

ccmlst (approved):

PR #567, opened 2024-09-03
fix(DBCluster): Handle delete handler stabilization when dbcluster listempty

angusy29:
*Description of changes:*

When a user deletes their DBCluster, the DBCluster is first deleted via DeleteDbCluster.
The handler then runs stabilization, where it calls describeDbCluster.

When the describeDbCluster occurs right after DeleteDbCluster, the DescibeDbCluster call can return an empty DbClusters list instead of a DBClusterNotFoundException, and this causes the DeleteHandler to fail.

If the user re-attempts a stack deletion, the second try of stack deletion will succeed, because the DescribeDbCluster call will throw a DBClusterNotFoundException.


aws-rds-dbcluster/src/main/java/software/amazon/rds/dbcluster/BaseHandlerStd.java
```diff
@@ -401,7 +401,7 @@ protected boolean isDBClusterDeleted(
     ) {
         try {
             fetchDBCluster(proxyClient, model);
-        } catch (DbClusterNotFoundException e) {
+        } catch (DbClusterNotFoundException | IndexOutOfBoundsException e) {
```
angusy29, line 404: I thought about making this catch at fetchDBCluster (further up the call stack), but `isDBClusterDeleted` is only called in 1 callsite and contains the blast radius to just the DeleteHandler.

ccmlst (approved):

PR #572, opened 2024-10-22
[DBInstance] Support for Oracle Multitenant SID

angusy29:
*Description of changes:*

DBSystemId is a property used to customise RDS Oracle Custom container databases' system id.

This is a createOnlyProperty which requires replacement of the DBInstance resource if the customer attempts to rename it. 

Because DBSystemId is only passed in on [create-db-instance](https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-instance.html), it means that there is potential for open fail scenarios, where user sets the DBSystemId and sees no results.

For example: We want to block users from being able to set a DBSystemId for [read replicas](https://docs.aws.amazon.com/cli/latest/reference/rds/create-db-instance-read-replica.html), [restore from PITR](https://docs.aws.amazon.com/cli/latest/reference/rds/restore-db-instance-to-point-in-time.html) and [restore from snapshot](https://docs.aws.amazon.com/cli/latest/reference/rds/restore-db-instance-from-db-snapshot.html)

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/validators/OracleCustomSystemId.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/validators/OracleCustomSystemId.java
@@ -1,0 +1,21 @@
+package software.amazon.rds.dbinstance.validators;
+
+import software.amazon.rds.common.request.RequestValidationException;
+import software.amazon.rds.dbinstance.ResourceModel;
+import software.amazon.rds.dbinstance.util.ResourceModelHelper;
+
+public class OracleCustomSystemId {
+    public static void validateRequest(final ResourceModel model) throws RequestValidationException {
+        if (ResourceModelHelper.isRestoreToPointInTime(model) && model.getDBSystemId() != null) {
+            throw new RequestValidationException("Cannot set DBSystemId when restoring to point in time");
+        }
+
+        if (ResourceModelHelper.isDBInstanceReadReplica(model) && model.getDBSystemId() != null) {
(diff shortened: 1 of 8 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/validators/OracleCustomSystemId.java
```diff
@@ -0,0 +1,21 @@
+package software.amazon.rds.dbinstance.validators;
+
+import software.amazon.rds.common.request.RequestValidationException;
+import software.amazon.rds.dbinstance.ResourceModel;
+import software.amazon.rds.dbinstance.util.ResourceModelHelper;
+
+public class OracleCustomSystemId {
+    public static void validateRequest(final ResourceModel model) throws RequestValidationException {
+        if (ResourceModelHelper.isRestoreToPointInTime(model) && model.getDBSystemId() != null) {
```
parker4cpk, line 9: You'll probably have to add tests where `model.getDBSystemId()` resolves to `false` in order to pass the coverage.

parker4cpk (approved):

aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/validators/OracleCustomSystemId.java
ccmlst, line 9: Consider short circuiting the logic by return early when DBSystemID is null and remove all the "&&" condition. This simplifies the code and make it easier for others to read.

PR #574, opened 2024-11-04
[CFN-27] Working towards a more extensible DBInstance create flow....

npxcomplete:
Incomplete, for early feedback from the other core maintainers. We (AWS RDS) have a number of upcoming features that stand to benefit from presenting an cleaner way to extend the create paths. The core team have discussed what this might look like and this is a first pass attempt to separate shape the code to that vision. 

*Description of changes:* Static methods and error rules have been pulled out of the base handler class; The instance create factory is now constructed by factory according to the model. 

I plan to follow up with a second revision but wanted to stop here as the decomposition of the BaseHandlerStd, while simple, has become somewhat large. The "FreshInstance" factory has copied a few methods that need to be reconciled. 

### DIFF
--- a/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
+++ b/aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
@@ -13,83 +13,37 @@ import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apache.commons.lang3.BooleanUtils;
 
-import com.amazonaws.arn.Arn;
-import com.amazonaws.util.CollectionUtils;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 import software.amazon.awssdk.services.ec2.Ec2Client;
 import software.amazon.awssdk.services.ec2.model.DescribeSecurityGroupsResponse;
 import software.amazon.awssdk.services.ec2.model.SecurityGroup;
 import software.amazon.awssdk.services.rds.RdsClient;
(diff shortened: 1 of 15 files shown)


aws-rds-dbinstance/src/main/java/software/amazon/rds/dbinstance/BaseHandlerStd.java
```diff
@@ -715,150 +393,48 @@ private void resourceStabilizationTime(final CallbackContext context) {
...
-    }
-
-    boolean isReplicationComplete(final DBInstance dbInstance) {
-        return Optional.ofNullable(dbInstance.statusInfos()).orElse(Collections.emptyList())
-                .stream()
-                .filter(statusInfo -> READ_REPLICA_STATUS_TYPE.equals(statusInfo.statusType()))
-                .allMatch(statusInfo -> ReadReplicaStatus.Replicating.equalsString(statusInfo.status()));
+        final Optional<DBCluster>  maybeDbCluster = DBInstancePredicates.isDBClusterMember(model)
+            ? Optional.of(fetchDBCluster(rdsProxyClient, model))
+            : Optional.empty();
```
angusy29, line 414: Can we not use optionals, it adds more cognitive load for the reader, can we instead use method overloads for when there is a dbcluster vs when there isn't?

From Brian Goetz himself: https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555

And: http://dolszewski.com/java/java-8-optional-use-cases/

npxcomplete: We can not-use Optionals here. But let's avoid taking the java authors word as gospel. What they intended Optionals for says nothing about what they are good for. What they are good for is avoiding `null`. The cognitive load of an optional is really just the cognitive load of a nullable reference, but it's in your face so you/we are more likely to handle it correctly.

PR #585, opened 2024-12-02
add global cluster endpoint to rds-globalcluster

akwan224:
*Description of changes:*
Following up with the launch of global cluster endpoints https://aws.amazon.com/about-aws/whats-new/2024/10/amazon-aurora-global-database-writer-endpoint, adding global cluster endpoints to rds-global-cluster 


aws-rds-globalcluster/aws-rds-globalcluster.json
```diff
@@ -41,9 +41,19 @@
         }
       ]
     },
-    "StorageEncrypted": {
+    "StorageEncrypted": { 
       "description": " The storage encryption setting for the new global database cluster.\nIf you specify the SourceDBClusterIdentifier property, don't specify this property. The value is inherited from the cluster.",
       "type": "boolean"
+    },
+    "GlobalEndpoint": {
```
ccmlst, line 48: how is this property gets used?

akwan224: not passed in as part of the createGlobalCluster request, it is created with the global cluster

Dmytrok-amz: It seems to me we need to add it as a readonly property (similar to e.g. https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-rds/blob/master/aws-rds-dbcluster/aws-rds-dbcluster.json#L436) - see https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html 

readOnlyProperties
    Resource properties that can be returned by a read or list request, but can't be set by the user.

PR #592, opened 2025-02-14
[DBShardGroup] Introduce tag-on-create behavior

ccmlst:
*Description of changes:*

[DBShardGroup] Introduce tag-on-create behavior


ccmlst:
Checks failed because Python 3.7 has reached the EOL and it's not supported in the Ubuntu 24.04.1. I will create another PR to fix build failure then recreate this PR. 
See: 
https://github.com/actions/setup-python/issues/962#issuecomment-2414418045
https://devguide.python.org/versions/

